PR #54, opened 2020-10-21
Updates OAuth to use OkHttp.

codingjeremy:
This is more a swap out to OkHttp then OAuth changes.


JohnNichol (approved):
Thanks Jeremy looks good

PR #55, opened 2020-10-21
Deprecate several older samples.

codingjeremy:
Pretty easy review... just removing 4 older samples that I verified we should deprecate.


JohnNichol (approved):
All pretty straightforward - thanks

PR #69, opened 2021-01-14
Update Always On Sample

JolandaVerhoef:
Readme was still pointing to the old support library classes, although the actual code had been updated already. Build gradle contained some unnecessary parts and old dependencies.

Code didn't run on my machine because `@color/green` could not be resolved. Added the color.


AlwaysOn/README.md
```diff
@@ -7,26 +7,25 @@ A basic sample showing how to support ambient mode for native Wear apps.
 Introduction
 ------------
 
-[WearableActivity][1] offers methods for supporting your native app staying on the screen when the Wear device enters ambient mode.
```
codingjeremy, line 10: Can we add this note somewhere? It's important as we don't want devs using always on unless they have a really strong use case for it. (I added it in the class, but seeing your edits, it should be in README too.)

IMPORTANT NOTE:  
Most apps shouldn't use the always on/ambient mode APIs, as it drains battery life and the system already handles this for you. 

That is, by default (without this API), the system will go into system ambient mode if the user hasn't interacted with the watch in a period of time, and if the user interacts again with the watch soon after that, the system will automatically bring up your app again in the same state it was before that point which covers most use cases.

codingjeremy (commented):
Overall, LGTM. 

I just wanted to add a note about not using it unless you have a strong use case. Have a look at my comment and edit it as needed. Thanks!

codingjeremy (approved):

PR #71, opened 2021-02-01
Replacing usage of PhoneDevceType from WSL with new PhoneTypeHelper from AndroidX

NedaTop:


WearVerifyRemoteApp/Wearable/build.gradle
```diff
@@ -22,6 +22,7 @@ repositories {
 dependencies {
 
     implementation 'androidx.wear:wear:1.1.0'
+    implementation 'androidx.wear:wear-phone-interactions:1.0.0-alpha01'
```
castedmo, line 25: This library includes a AIDL (shared with WSL). Not sure if it can/should be mixed with WSL in dependencies.

PR #74, opened 2021-02-11
Bumps version and adds default client id field in strings.

codingjeremy:


WearStandaloneGoogleSignIn/app/src/main/res/values/strings.xml
```diff
@@ -19,4 +19,5 @@
     <string name="wear_signout_button_text">Sign out</string>
     <string name="google_signin_successful">Google Sign-In successful!</string>
     <string name="signout_successful">Successfully signed out!</string>
+    <string name="server_client_id">TODO_ADD_CLIENT_ID_XXXXXXX.apps.googleusercontent.com</string>
```
JolandaVerhoef, line 22: You added this string but you're not using it anywhere, correct?

codingjeremy: Correct, well, I'm not using it directly, but the google sign in API uses it.

codingjeremy: I added a comment to call that out.

JolandaVerhoef (approved):

PR #103, opened 2021-06-23
[Complications] Convert Java to Kotlin

alexvanyo:
The main conversion from Java to Kotlin for the `Complications` sample. The first commit contains the automatic conversion, and the others add some cleanup commits.

In particular, I replaced shared preferences with `DataStore`.


WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/DataStorage.kt
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+ * Updates the current state for a given complication, incrementing it by 1.
+ */
+suspend fun ComplicationToggleArgs.updateState(context: Context) {
+    val stateKey = getStatePreferenceKey()
+    context.dataStore.edit { preferences ->
+        val currentValue = preferences[stateKey] ?: 0
+        preferences[stateKey] = currentValue + 1
```
codingjeremy, line 47: super small nit (optional): I know the original just incremented forever... maybe we should put a module on the end so it never goes beyond the top integer value... maybe not... I will leave to you...

you could do % 100 or something..

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/IconProviderService.kt
```diff
@@ -0,0 +1,72 @@
...
+        val args = ComplicationToggleArgs(
+            providerComponent = ComponentName(this, javaClass),
+            complicationId = complicationId,
+        )
+        val complicationTogglePendingIntent =
+            ComplicationToggleReceiver.getComplicationToggleIntent(
+                context = this,
+                args = args,
+            )
+        val state = args.getState(this@IconProviderService)
```
codingjeremy, line 42: I would add a comment here (and the other Services where you do this about what is going on... It might not be clear right away why you creating a version of the sample that enables coroutines. Maybe something like this?

// Retrieves state from DataStore (suspend method).

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/IconProviderService.kt
```diff
@@ -0,0 +1,72 @@
...
+package com.example.android.wearable.wear.wearcomplicationproviderstestsuite
+
+import android.content.ComponentName
+import android.graphics.drawable.Icon
+import android.support.wearable.complications.ComplicationData
+import android.support.wearable.complications.ComplicationManager
+
+/**
+ * A complication provider that supports only [ComplicationData.TYPE_ICON] and cycles through
+ * a few different icons on each tap.
```
codingjeremy, line 25: It's important you call out why you are using a custom version of this class. I would add an additional comment here and the other files... maybe something like this...

Note: Uses a custom version of ComplicationProviderService to support coroutines, so data operations (specifically, calls to DataStore) can be supported directly in the onComplication update. If you don't have any of these operations, you can use the normal approach.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/IconProviderService.kt
LouisCAD: Nit on the review comment: "suspend **function**" is more accurate than "suspend _method_"

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/DataStorage.kt
alexvanyo: Updated in latest commit. I added a comment here acknowledging the possible overflow, and I updated the logic in the different services to avoid any crashes. Upon overflowing, the toggling order might change, but since I updated the state to a `Long` that would require a few quintillion taps.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/IconProviderService.kt
alexvanyo: Added in the latest commit! I also made `NoDataProviderService` subclass `ComplicationProviderService` directly since it doesn't call any suspending functions, and linked to it from the other services.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/IconProviderService.kt
alexvanyo: Added!

(It's a shame that the "suspension point" gutter icons only show up in the IDE, since they are really helpful to see)

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/ComplicationToggleReceiver.kt
```diff
@@ -0,0 +1,77 @@
...
+    companion object {
+        private const val EXTRA_ARGS = "arguments"
+
+        /**
+         * Returns a pending intent, suitable for use as a tap intent, that causes a complication to be
+         * toggled and updated.
+         */
+        fun getComplicationToggleIntent(
+            context: Context,
+            args: ComplicationToggleArgs,
```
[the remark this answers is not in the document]
alexvanyo: Added a commit to remove all of the trailing commas I added.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/LongTextProviderService.kt
```diff
@@ -19,11 +19,21 @@ import android.content.ComponentName
...
 import android.support.wearable.complications.ComplicationManager
+import android.support.wearable.complications.ComplicationProviderService
 import android.support.wearable.complications.ComplicationText
+import androidx.datastore.core.DataStore
 
 /**
  * A complication provider that supports only [ComplicationData.TYPE_LONG_TEXT] and cycles
  * through the possible configurations on tap.
+ *
+ * Note: This subclasses [ComplicationProviderService] to support coroutines, so data operations
```
codingjeremy, line 30: Maybe reword? Technically, this class doesn't subclass ComplicationProviderService. Maybe...

Note: [SuspendingComplicationProviderService] adds coroutine support to [ComplicationProviderService] so data operations (specifically, calls to [DataStore]) can be supported directly in the [onComplicationUpdate].
 See [SuspendingComplicationProviderService] for the implementation details... 

then keep the rest?

codingjeremy (approved):
One nit on the description. Outside of that, LGTM.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/LongTextProviderService.kt
alexvanyo: Whoops, that was a typo! Reworded in the latest commit.

codingjeremy (approved):

PR #109, opened 2021-06-30
[Complications] Convert to androidx wear libraries

alexvanyo:
Migrates complications to use the latest `androidx:wear` library (alpha17) from the support library.


codingjeremy (approved):
LGTM, adding @ADClarke to do a final check.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/ComplicationToggleArgs.kt
```diff
@@ -27,11 +27,11 @@ import kotlinx.parcelize.Parcelize
...
-    val complicationId: Int
+    val complicationInstanceId: Int
 ) : Parcelable
 
 /**
  * Returns the key for the preference used to hold the current state of a given complication.
  */
 fun ComplicationToggleArgs.getStatePreferenceKey(): Preferences.Key<Long> =
-    longPreferencesKey("${providerComponent.className}$complicationId")
+    longPreferencesKey("${providerComponent.className}$complicationInstanceId")
```
LouisCAD, line 37: This is unlikely to be stable across minified builds since R8 will obfuscate the class names differently, leading to the terrible user experience of preferences reset on each app update. Best to not show such an example to developers.

LouisCAD (changes requested):

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/ComplicationToggleArgs.kt
ADClarke: Agreed, although you could consider turning obfuscation off for these classes.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/LongTextProviderService.kt
```diff
@@ -53,52 +59,138 @@ class LongTextProviderService : SuspendingComplicationProviderService() {
...
+        val title: ComplicationText?
+        val monochromaticImage: MonochromaticImage?
+        val contentDescription: ComplicationText
+
+        when (case) {
+            Case.TEXT_ONLY -> {
+                text = PlainComplicationText.Builder(
+                    text = getText(R.string.long_text_only)
+                ).build()
+                smallImage = null
```
ADClarke, line 90: Kotlin tends to embrace terseness. In that spirit I'd suggest doing something like this:

private fun getComplicationData(tapAction: PendingIntent?, case: Case) = when (case) {
  Case.TEXT_ONLY -> LongTextComplicationData.Builder(
            text = PlainComplicationText.Builder(
                    text = getText(R.string.long_text_only)
                ).build(),
            contentDescription = PlainComplicationText.Builder(
                    text = getText(R.string.long_text_only_content_description)
                ).build()
        ).build()

  Case.TEXT_WITH_ICON -> etc...
}

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/RangedValueProviderService.kt
```diff
@@ -54,46 +59,115 @@ class RangedValueProviderService : SuspendingComplicationProviderService() {
...
+        val caseContentDescription: String
+
+        val minValue = case.minValue
+        val maxValue = case.maxValue
         val value = Random.nextDouble(minValue.toDouble(), maxValue.toDouble()).toFloat()
+        val percentage = (value - minValue) / (maxValue - minValue)
+
+        when (case) {
+            Case.TEXT_ONLY -> {
+                text = PlainComplicationText.Builder(
```
ADClarke, line 92: Ditto here.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/ShortTextProviderService.kt
```diff
@@ -53,41 +57,95 @@ class ShortTextProviderService : SuspendingComplicationProviderService() {
...
+        tapAction: PendingIntent?,
+        case: Case
+    ): ComplicationData {
+        val text: ComplicationText
+        val title: ComplicationText?
+        val monochromaticImage: MonochromaticImage?
+        val contentDescription: ComplicationText
+
+        when (case) {
+            Case.TEXT_ONLY -> {
```
ADClarke, line 83: Ditto.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/SmallImageProviderService.kt
```diff
@@ -52,28 +58,61 @@ class SmallImageProviderService : SuspendingComplicationProviderService() {
...
+
+    private fun getComplicationData(
+        tapAction: PendingIntent?,
+        case: Case
+    ): ComplicationData {
+        val smallImage: SmallImage
+        val contentDescription: ComplicationText
+
+        when (case) {
+            Case.PHOTO -> {
```
ADClarke, line 82: Ditto.

ADClarke (commented):
Generally looks fine but a few comments mostly to make the code shorter.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/ComplicationToggleArgs.kt
LouisCAD: Then later on, someone enables minification again because R8 optimizations are wanted, or even needed (e.g. for Compose efficiency and jank-lessness), and it breaks.

IMO, there should be a lint rule with error level for these risky key naming schemes.

alexvanyo: In this instance, I believe this usage _should_ be fine since the class names are referring to `Service`s, which won't be obfuscated. These names have to be stable, since if they change, the complication will disappear from the watchface if the user was using it (very similar to widgets or wallpapers).

That being said, I totally agree that this usage is risky. I'll update it with a more straightfoward approach.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/LongTextProviderService.kt
alexvanyo: Updated most of the builders to follow this pattern in 629b4e8!

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/RangedValueProviderService.kt
alexvanyo: I left `RangedValueProviderService` as is, due to the `contentDescription` logic that is dependent on the case.

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/ComplicationToggleArgs.kt
alexvanyo: Updated to use explicit keys in 8073a9a

WearComplicationProvidersTestSuite/Wearable/src/main/java/com/example/android/wearable/wear/wearcomplicationproviderstestsuite/Complications.kt
```diff
@@ -0,0 +1,19 @@
...
+     */
+    val key: String
+) {
+    ICON("Icon"),
+    LARGE_IMAGE("LargeImage"),
+    LONG_TEXT("LongText"),
+    RANGED_VALUE("RangedValue"),
+    SHORT_TEXT("ShortText"),
+    SMALL_IMAGE("SmallImage")
+}
```
LouisCAD, line 19: I think it'd make more sense to drop SCREAMING_CASE for PascalCase and use the built-in `name` property of enums in place of the `key` property and the repeated text.

alexvanyo: I don't have too strong of a preference, but I do like the extra explicitness of a separate `key`.

alexvanyo: Definitely true, although it's also a bit magical that `name`s of enum cases don't get obfuscated either by Proguard, even though the values themselves are.

On the duplicate note, I'm planning to add tests for these keys, so I will definitely be ensuring that the keys are unique there.

PR #110, opened 2021-06-30
[Tiles] Update to alpha08

alexvanyo:
Updates the tiles library to the latest version.

One interesting note:
Updating without bumping `kotlinx-coroutines-guava` caused a runtime crash in the renderer along the lines of

```
java.lang.NoSuchMethodError: No direct method <init>(Lkotlin/coroutines/CoroutineContext;ZILkotlin/jvm/internal/DefaultConstructorMarker;)V in class Lkotlinx/coroutines/AbstractCoroutine; or its super classes (declaration of 'kotlinx.coroutines.AbstractCoroutine' appears in /data/app/...
```

I'm guessing this was due to the version bump of `kotlinx.coroutines` in `androidx` to `1.5.0`. This bumped `kotlinx-coroutines-android` and `kotlinx-coroutines-core` to `1.5.0`, but left `kotlinx-coroutines-guava` on `1.4.2`.

Updated the Kotlin version and `kotlinx-coroutines` version to latest.

### DIFF
--- a/WearTilesKotlin/app/build.gradle
+++ b/WearTilesKotlin/app/build.gradle
@@ -46,14 +46,13 @@ android {
 }
 
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.4.3'
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-guava:1.4.2'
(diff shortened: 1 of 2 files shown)


WearTilesKotlin/app/build.gradle
```diff
@@ -46,14 +46,13 @@ android {
 }
 
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
alexvanyo, line 49: The standard library is depended on by default since 1.4, so removed the explicit dependency here.

codingjeremy (approved):

PR #111, opened 2021-07-01
[Tiles] Add content descriptions

alexvanyo:
Adds content descriptions to the three sample tiles. I also made all buttons actually clickable with a no-op `LoadAction`.

### DIFF
--- a/WearTilesKotlin/app/src/main/java/com/example/wear/tiles/fitness/FitnessTileService.kt
+++ b/WearTilesKotlin/app/src/main/java/com/example/wear/tiles/fitness/FitnessTileService.kt
@@ -18,20 +18,19 @@ package com.example.wear.tiles.fitness
 import androidx.core.content.ContextCompat
 import androidx.wear.tiles.TileProviderService
 import androidx.wear.tiles.ColorBuilders.argb
-import androidx.wear.tiles.DeviceParametersBuilders
 import androidx.wear.tiles.DeviceParametersBuilders.DeviceParameters
 import androidx.wear.tiles.DimensionBuilders.degrees
 import androidx.wear.tiles.DimensionBuilders.dp
-import androidx.wear.tiles.DimensionBuilders.sp
 import androidx.wear.tiles.LayoutElementBuilders.ARC_ANCHOR_START
 import androidx.wear.tiles.LayoutElementBuilders.Arc
 import androidx.wear.tiles.LayoutElementBuilders.ArcLine
 import androidx.wear.tiles.LayoutElementBuilders.Box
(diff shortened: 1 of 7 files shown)


codingjeremy (approved):
@jamie-garside mind taking a final look?

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/fitness/FitnessTileService.kt
```diff
@@ -118,18 +118,25 @@ class FitnessTileService : TileProviderService() {
...
-                            .setText(resources.getString(R.string.goal, goalProgress.goal))
-                            .setFontStyle(
-                                FontStyle.builder().setSize(sp(44f)).build()
-                            )
+                            .setText(getString(R.string.tile_fitness_goal, goalProgress.goal))
                             .setFontStyle(FontStyles.title3(deviceParameters))
                     )
             )
+            .setModifiers(
+                ModifiersBuilders.Modifiers.builder()
```
jamie-garside, line 130: Nit: import Modifiers and Semantics directly to keep this consistent with the other builders.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -190,39 +190,87 @@ class MessagingTileService : TileProviderService() {
...
+    private fun contactLayout(
+        contact: Contact,
+        deviceParameters: DeviceParameters,
+        clickable: Clickable
+    ) = Box.builder().apply {
+        val modifiersBuilder = Modifiers.builder()
+            .setClickable(clickable)
+            .setSemantics(
+                ModifiersBuilders.Semantics.builder()
+                    .setContentDescription(contact.initials)
```
jamie-garside, line 255: Since this is a content description instead, should we use the full name rather than initials?

jamie-garside (approved):
Looks good, thanks!

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/fitness/FitnessTileService.kt
alexvanyo: Updated in the latest commit!

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
alexvanyo: Updated in the latest commit!

PR #119, opened 2021-07-13
[VerifyRemote] Convert wear app to Kotlin

alexvanyo:
Converts the wear portion of the VerifyRemote app to Kotlin, and modernizes it.

Highlights:
- Updates to use `ViewBinding`
- Converts to coroutines
- Converts to the `androidx.wear` library
- Cleans up the project
- Adjusts the layout to be scrollable


WearVerifyRemoteApp/Wearable/src/main/java/com/example/android/wearable/wear/wearverifyremoteapp/MainWearActivity.kt
```diff
@@ -0,0 +1,206 @@
...
+        nodeClient = Wearable.getNodeClient(this)
+        remoteIntentHelper = RemoteIntentHelper(this)
+
+        binding.informationTextView.text = getString(R.string.message_checking)
+        binding.remoteOpenButton.setOnClickListener { 
+            openAppInStoreOnPhone()
+        }
+
+        if (resources.configuration.isScreenRound) {
+            binding.scrollingContentContainer.doOnPreDraw {
```
jdkoren, line 77: doOnNextLayout?

WearVerifyRemoteApp/Wearable/src/main/java/com/example/android/wearable/wear/wearverifyremoteapp/MainWearActivity.kt
```diff
@@ -0,0 +1,206 @@
...
+                // Calculate the padding necessary to make the scrolling content fit in a square inscribed on a round
+                // screen.
+                it.setPadding((it.width / 2.0 * (1.0 - 1.0 / sqrt(2.0))).roundToInt())
+            }
+        }
+    }
+
+    override fun onPause() {
+        Log.d(TAG, "onPause()")
```
jdkoren, line 86: I would remove Log messages that don't print something related to the sample (unless the sample is demonstrating a certain timing or execution, but I don't get that impression here).

WearVerifyRemoteApp/Wearable/src/main/java/com/example/android/wearable/wear/wearverifyremoteapp/MainWearActivity.kt
alexvanyo: `doOnLayout` and `doOnNextLayout` let the default/0 padding be drawn for one frame, which `doOnPreDraw` avoids.

WearVerifyRemoteApp/Wearable/src/main/java/com/example/android/wearable/wear/wearverifyremoteapp/MainWearActivity.kt
alexvanyo: That makes sense to me, I left them in since they were there previously, took out the lifecycle ones with 5e65373.

WearVerifyRemoteApp/Wearable/src/main/java/com/example/android/wearable/wear/wearverifyremoteapp/MainWearActivity.kt
jdkoren: Fascinating, and unexpected considering that padding can affect layout, and layout should happen before draw...

jdkoren (approved):

WearVerifyRemoteApp/Wearable/src/main/java/com/example/android/wearable/wear/wearverifyremoteapp/MainWearActivity.kt
alexvanyo: Just dived into the source for this more, I was curious what the exact difference was.

You are absolutely correct that layout will happen before draw, but `doOnPreDraw` has the special behavior of [cancelling the draw call](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ViewTreeObserver.java;l=1085-1100) for the layout pass without the padding, so the user only sees the layout with the padding.

The most ideal solution to this would probably be a subclass of the layout that does all of this in `onMeasure` to avoid the double layout pass (a la `BoxInsetLayout`), but that seems like a bit of overkill here.

alexvanyo:
Thanks @jdkoren !

PR #120, opened 2021-07-13
Updates watch face to latest alpha and removes data layer.

codingjeremy:
Refactors alpha samples away from using its own Room database layer while also updating to the latest APIs.

Values needed to render watch face are mostly constants. Values that can change, are stored in the watch face apis, we just listen to changes using a flow and update renderer that way.

Followup CL will include Activity to edit watch face values.


WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/WatchFaceUserSettingChangesDataSource.kt
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Data Source for all changes by the user in settings related to the watch face styles. Coverts
+ * the watch face api callback to a flow.
+ */
+class WatchFaceUserSettingChangesDataSource(
+    private val currentUserStyleRepository: CurrentUserStyleRepository
+) {
+    @ExperimentalCoroutinesApi
+    fun getUserWatchFaceChanges(): Flow<WatchFaceUserChanges> = callbackFlow {
+        val userStyleChangeListener = object : CurrentUserStyleRepository.UserStyleChangeListener {
```
alexvanyo, line 42: I think this could be a bit clearer if the conversion from a callback to a `Flow` happens as soon as possible. Or in other words, first constructing a `Flow<UserStyle>` directly from the callback, and then using `map` or other operators to transform the `Flow`.

That'd keep the `callbackFlow` section short.

Wrapping that into a helper of
```kotlin
val CurrentUserStyleRepository.userStyleChanges: Flow<UserStyle>
```
might be a good `ktx` candidate.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/WatchFaceUserSettingChangesDataSource.kt
```diff
@@ -0,0 +1,107 @@
...
+                        // TODO (codingjeremy): Add complication change support if settings activity
+                        // PR doesn't cover it. Otherwise, remove comment.
+                    }
+                }
+
+                try {
+                    trySend(newWatchFaceUserChangesChanges).isSuccess
+                } catch (error: Throwable) {
+                    Log.d(TAG, "error sending Flow: $error")
+                }
```
alexvanyo, line 83: Based on the usage and `UserStyleChangeListener` docs, I believe we only care about the most recent `UserStyle` since it is a complete snapshot (as opposed to partial updates).

If that's correct, we can add a `.buffer(Channel.CONFLATED)` right after the `callbackFlow`, which overrides the default, limited buffer size. Then `trySend` will only fail if the channel is closed, which indicates that we're being cancelled, so we can drop the `catch` block here.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/viewmodels/AnalogWatchFaceViewModel.kt
```diff
@@ -15,123 +15,52 @@
...
-        val newHoursPipsSettingStringVersion: String =
-            userStyle.toMap()[DRAW_HOUR_PIPS_STYLE_SETTING] ?: ""
-
-        val newHoursPipsSetting =
-            if (newHoursPipsSettingStringVersion.isEmpty()) {
-                revisedAnalogWatchFaceEntity.drawHourPips
-            } else {
-                newHoursPipsSettingStringVersion.toBoolean()
+    init {
+        scope.launch {
```
alexvanyo, line 48: This would be a great place for a more declarative version with `stateIn`:

```kotlin
    val uiState: StateFlow<UserChangesUiState> =
        watchFaceUserSettingChangesDataSource.getUserWatchFaceChanges()
            .map(UserChangesUiState::Success)
            .stateIn(
                scope = scope,
                started = SharingStarted.Eagerly,
                initialValue = UserChangesUiState.Success(WatchFaceUserChanges())
            )
```

`SharingStarted.Eagerly` would be the equivalent replacement, but `SharingStarted.WhileSubscribed` might be good as well to avoid doing any unnecessary work if nobody is listening.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/viewmodels/AnalogWatchFaceViewModel.kt
```diff
@@ -15,123 +15,52 @@
...
 /**
- * View model used for [AnalogWatchFaceService].
+ * View model used for [AnalogWatchFaceService]. Uses flow to receive changes in the data source and
+ * because the renderer is a service without LifeCycle support, handles canceling scope when no
+ * longer needed.
  */
-class AnalogWatchFaceViewModel(private val repository: WatchFaceRepository) {
-
-    // Used to launch coroutines (non-blocking way to insert data).
+@ExperimentalCoroutinesApi
```
alexvanyo, line 36: Nit: I like using the `@OptIn` version for these to clearly show the place where we are choosing to use an experimental piece of the API, which stops the propagating of `@ExperimentalCoroutinesApi` for any class that uses it.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/WatchFaceUserSettingChangesDataSource.kt
codingjeremy: You are correct, I only care about the most recent version. 

For some reason, I'm getting an error when I add .buffer() after my callbackFlow, so I'll fix this in a followup commit tomorrow when I figure out what the issue is.

codingjeremy (commented):
Addressed most of the comments, PTAL.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/watchface/ColorStyleIdAndResourceIds.kt
```diff
@@ -0,0 +1,129 @@
...
+ * Represents the id (in the watch face database) and the resource ids for the color styles in the
+ * watch face (including the complication color style).
+ *
+ * The watch face renderer (with the context), will translate these into the actual colors and
+ * ComplicationDrawables to use for rendering.
+ */
+sealed class ColorStyleIdAndResourceIds(
+    val id: String,
+    val name: Int,
+    val iconResourceId: Int,
```
alexvanyo, line 54: `name` and `iconResourceId` are both coming up as unused, so they should probably be removed, or it looks they should be read in `UserStyleSettingUtils.kt`?

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/watchface/ColorStyleIdAndResourceIds.kt
```diff
@@ -0,0 +1,129 @@
...
+        iconResourceId = WHITE_COLOR_STYLE_ICON_ID,
+        complicationStyleDrawableId = R.drawable.complication_white_style,
+        primaryColorId = R.color.white_primary_color,
+        secondaryColorId = R.color.white_secondary_color,
+        backgroundColorId = R.color.white_background_color,
+        outerElementColorId = R.color.white_outer_element_color
+    )
+
+    companion object {
+        // Translates the string id to the correct ColorStyleIdAndResourceIds object.
```
alexvanyo, line 117: Nit: Here and a few other places, the `//` comment won't show up in the documentation for the function: 
```suggestion
        /**
         * Translates the string id to the correct ColorStyleIdAndResourceIds object.
         */
```

WatchFaceAlphaKotlin/build.gradle
```diff
@@ -17,26 +17,26 @@
...
         activityVersion = '1.1.0'
         appCompatVersion = '1.2.0'
         constraintLayoutVersion = '2.0.2'
         coreTestingVersion = '2.1.0'
         coroutines = '1.4.2'
-        lifecycleVersion = '2.2.0'
+        lifecycleVersion = '2.3.1'
         materialVersion = '1.2.1'
-        roomVersion = '2.2.5'
+        roomVersion = '2.3.0'
```
alexvanyo, line 30: Nit: There's a few versions here that are no longer being used that'd be good to clean up.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/watchface/ColorStyleIdAndResourceIds.kt
codingjeremy: I reworked the settings code to use these instead and added a translation function to covert them. PTAL.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/watchface/ColorStyleIdAndResourceIds.kt
codingjeremy: Addressed

WatchFaceAlphaKotlin/build.gradle
codingjeremy: Done.

codingjeremy (commented):
Addressed all comments, PTAL.

ADClarke (commented):
I've mostly looked at the service class and passing the UserStyle down to the renderer.  Those bits LGTM thanks.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/watchface/ColorStyleIdAndResourceIds.kt
```diff
@@ -0,0 +1,167 @@
...
+         *
+         * IMPORTANT: To avoid reflection, the full list of ColorStyleIdAndResourceIds are manually
+         * included below (you need to use reflection to get full list of sealed classes which I did
+         * not want to do). If you add a new style, add it to the list below. Otherwise, it will not
+         * show up to the user!
+         */
+        fun toOptionList(context: Context): List<ListUserStyleSetting.ListOption> {
+            // Hardcoded list of all ColorStyleIdAndResourceIds. If you add a new one, add it here
+            // as well.
+            val colorStyleIdAndResourceIdsList = listOf(Red, Green, Blue, White, Ambient)
```
alexvanyo, line 148: Could we make `ColorStyleIdAndResourceIds` an `enum class`? Then we could use `values()` or `enumValues()`, to avoid having to remember to update this list.

I am also quite partial to a certain library that allows generating these lists of `sealed class`s with only `object`s 😸

WatchFaceAlphaKotlin/app/build.gradle
```diff
@@ -54,34 +54,25 @@ android {
...
-    implementation "androidx.wear:wear-watchface:$rootProject.watchFaceAlphaVersion"
 
-    implementation "androidx.wear:wear-watchface-client:$rootProject.watchFaceAlphaVersion"
-    implementation "androidx.wear:wear-watchface-complications-rendering:$rootProject.watchFaceAlphaVersion"
+    // Wear libraries
+    implementation 'androidx.wear:wear:1.1.0'
 
+    // Watch face specific libraries
     implementation "androidx.wear:wear-complications-data:$rootProject.watchFaceAlphaVersion"
     implementation "androidx.wear:wear-complications-provider:$rootProject.watchFaceAlphaVersion"
```
alexvanyo, line 70: Don't want to delay this PR, but the latest version of `complications` changed this dependency to `wear-complications-data-source`, which will require some API updates as a follow-up.

WatchFaceAlphaKotlin/app/src/main/AndroidManifest.xml
```diff
@@ -1,43 +1,50 @@
...
-        android:name=".MainApplication"
         android:allowBackup="true"
         android:fullBackupContent="@xml/my_backup_rules"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:supportsRtl="true"
         android:theme="@android:style/Theme.DeviceDefault">
+        <uses-library
+            android:name="com.google.android.wearable"
+            android:required="true" />
```
alexvanyo, line 38: Optional: This should be included automatically with the latest `androidx.wear:wear` library.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/WatchFaceUserSettingChangesDataSource.kt
alexvanyo: This looks good to me!

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/AnalogWatchFaceService.kt
```diff
@@ -15,95 +15,82 @@
...
+    ): ComplicationSlotsManager = createComplicationSlotManager(
+        context = applicationContext,
+        currentUserStyleRepository = currentUserStyleRepository
+    )
 
     override fun onDestroy() {
+        Log.d(TAG, "onDestroy()")
+
+        // Cancels scope used for retrieving watch face style changes via callbackFlow.
         analogWatchFaceViewModel.clear()
```
alexvanyo, line 58: Since `analogWatchFaceViewModel` is now initialized in `createWatchFace` instead of `onCreate`, is it guaranteed to be initialized in `onDestroy`? We might want to wrap this in `if (::analogWatchFaceViewModel.isInitialized) {`.

WatchFaceAlphaKotlin/app/src/main/AndroidManifest.xml
```diff
@@ -1,43 +1,50 @@
...
+    <uses-permission android:name="com.google.android.permission.PROVIDE_BACKGROUND" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
     <uses-permission android:name="com.google.android.wearable.permission.RECEIVE_COMPLICATION_DATA" />
     <uses-permission android:name="com.google.android.wearable.permission.RECEIVE_COMPLICATION_DATA_PRIVILEGED" />
     <uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
 
     <application
-        android:name=".MainApplication"
         android:allowBackup="true"
         android:fullBackupContent="@xml/my_backup_rules"
```
alexvanyo, line 31: Is the `my_backup_rules` still relevant, after the removal of the Room database?

WatchFaceAlphaKotlin/app/build.gradle
codingjeremy: Yes, I will update that in a follow up PR to get this one live. Should be easier to review then

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/watchface/ColorStyleIdAndResourceIds.kt
```diff
@@ -0,0 +1,167 @@
...
+                        id = UserStyleSetting.Option.Id(colorStyleIdAndResourceIds.id),
+                        displayName = context.getString(colorStyleIdAndResourceIds.name),
+                        icon = Icon.createWithResource(
+                            context,
+                            colorStyleIdAndResourceIds.iconResourceId
+                        )
+                    )
+                )
+            }
+            return result
```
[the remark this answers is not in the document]
codingjeremy: Done.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/watchface/ColorStyleIdAndResourceIds.kt
codingjeremy: Yeah, that makes sense.  Good bye sealed class... :'(

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/watchface/ColorStyleIdAndResourceIds.kt
```diff
@@ -0,0 +1,157 @@
...
+ */
+enum class ColorStyleIdAndResourceIds(
+    val id: String,
+    val nameResourceId: Int,
+    val iconResourceId: Int,
+    val complicationStyleDrawableId: Int,
+    val primaryColorId: Int,
+    val secondaryColorId: Int,
+    val backgroundColorId: Int,
+    val outerElementColorId: Int
```
alexvanyo, line 65: Nit: Could we add `@StringRes`, `@DrawableRes`, etc.?

alexvanyo (approved):
Looks good to me, just one last nit.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/watchface/ColorStyleIdAndResourceIds.kt
codingjeremy: done.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/WatchFaceUserSettingChangesDataSource.kt
```diff
@@ -0,0 +1,58 @@
...
+    @ExperimentalCoroutinesApi
+    fun getUserWatchFaceChanges(): Flow<UserStyle> = callbackFlow<UserStyle> {
+        val userStyleChangeListener = object : CurrentUserStyleRepository.UserStyleChangeListener {
+            override fun onUserStyleChanged(userStyle: UserStyle) {
+                Log.d(TAG, "onUserStyleChanged(): userStyle: $userStyle")
+                trySend(userStyle)
+            }
+        }
+
+        currentUserStyleRepository.addUserStyleChangeListener(userStyleChangeListener)
```
ADClarke, line 44: NB I think you can pass a lambda in here. E.g.

currentUserStyleRepository.addUserStyleChangeListener { userStyle ->
                Log.d(TAG, "onUserStyleChanged(): userStyle: $userStyle")
                trySend(userStyle)
}

ADClarke (approved):
Thanks

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/WatchFaceUserSettingChangesDataSource.kt
codingjeremy: Unfortunately, I can't use a SAM constructor without **fun** being in the interface name for Kotlin defined SAMs ([details](https://kotlinlang.org/docs/whatsnew14.html#sam-conversions-for-kotlin-interfaces)). 

However, it's a good idea, so I created a bug to add it to the API, and I can add this after. Thanks!

alexvanyo: We also want to maintain a reference here to allow removing the listener in `awaitClose`, so I don't think we'd want to use inline the listener either.

codingjeremy: If we use a SAM constructor (and the fun has been added to the interface), that should still work since we maintain a reference... unless I'm missing something?

```
        val userStyleChangeListener = CurrentUserStyleRepository.UserStyleChangeListener { userStyle ->
                Log.d(TAG, "onUserStyleChanged(): userStyle: $userStyle")
                trySend(userStyle)
        }
```

alexvanyo: Ah yes, that would still work and clean it up a bit. We wouldn't be able to go all the way though and do

```
currentUserStyleRepository.addUserStyleChangeListener { userStyle ->
```

PR #125, opened 2021-07-20
[RuntimePermissions] Add Kotlin plugin

alexvanyo:
Adds the Kotlin plugin to the `RuntimePermissions` sample in preparation for converting the sample from Java to Kotlin.


JolandaVerhoef (approved):
While you're at it, would you consider adding Spotless to these samples as well? Would be great to be able to run CI to check formatting & licenses etc in the future, like we do with the Compose samples.

alexvanyo:
@JolandaVerhoef great idea! I'll make a follow up PR for this sample and the other ones I've updated so far.

PR #129, opened 2021-07-26
[VerifyRemote] Watch app apply spotless checks

alexvanyo:
A small update to apply spotless to the `VerifyRemote` watch app.


JolandaVerhoef (approved):

JolandaVerhoef:
While we're at it, what do you think about adding Github Actions for this repo? Using compose-samples as an example, we can build the app, run spotlessCheck, and run tests if they are present.

alexvanyo:
@JolandaVerhoef That sounds great, I will do that!

PR #130, opened 2021-07-27
Add GitHub Actions CI

alexvanyo:
Adds GitHub Actions CI for most of the projects in the repo, based on the CI for https://github.com/android/compose-samples. This runs `./gradlew check`, which includes `lint`, `spotless`, and in the future, any non-instrumented tests.

I attempted to use [`android-emulator-runner`](https://github.com/ReactiveCircus/android-emulator-runner) to run `AlwaysOnKotlin`'s instrumented tests on CI, but I have this disabled for now due to flakiness.

### DIFF
--- a/AlwaysOnKotlin/Wearable/build.gradle
+++ b/AlwaysOnKotlin/Wearable/build.gradle
@@ -33,6 +33,7 @@ android {
 
     lintOptions {
         warningsAsErrors true
+        ignore "OldTargetApi" // API >30 doesn't exist for Wear OS
     }
 
     compileOptions {
@@ -60,9 +61,9 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.1"
-    implementation "androidx.activity:activity-ktx:1.2.4"
+    implementation "androidx.activity:activity-ktx:1.3.1"
     implementation "androidx.core:core-ktx:1.6.0"
-    implementation 'androidx.wear:wear:1.2.0-alpha12'
+    implementation 'androidx.wear:wear:1.2.0-alpha13'
 
     androidTestImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.5.1"
     androidTestImplementation "androidx.test:core:1.4.0"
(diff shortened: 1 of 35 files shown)


AlwaysOnKotlin/Wearable/build.gradle
```diff
@@ -20,13 +20,13 @@ plugins {
 }
 
 android {
-    compileSdkVersion 30
+    compileSdkVersion 31
```
codingjeremy, line 23: There is no Wear 31. :)

codingjeremy (commented):
I notice other sample repos have just one yml workflow for all the samples, that is, if one fails, the whole repo is red... IDK, should we have all these separated like this? Maybe so, I just wanted to verify :)

AlwaysOnKotlin/Wearable/build.gradle
alexvanyo: This is a curious point, I was going to ask about this. `lint`/Android Studio complains loudly if not targeting and compiling against the latest version, even though there currently isn't a Wear API 31. My guess would be that compiling and targeting against the latest API should be fine?

alexvanyo:
Per the multiple workflows, I was mostly basing this on the `compose-samples` set up, which also splits apart the workflows. One advantage is that we will only have to run the CI for the sample(s) that change. Another advantage is that each sample can be checked in parallel, without having to have a complicated single job that choreographs everything.

AlwaysOnKotlin/Wearable/build.gradle
codingjeremy: Hmm... that an interesting question... I'm not 100% sure there isn't something that could happen. Let's verify.

codingjeremy (commented):
I'm ok with the separate YML files per your details. 

Let's verify the 31 thing before moving forward.

scripts/checksum.sh
```diff
@@ -0,0 +1,40 @@
...
+FILES=()
+while read -r -d ''; do
+	FILES+=("$REPLY")
+done < <(find $SAMPLE -type f \( -name "build.gradle*" -o -name "gradle-wrapper.properties"  -o -name "robolectric.properties" \) -print0)
+
+# Loop through files and append MD5 to result file
+for FILE in ${FILES[@]}; do
+	echo $(checksum_file $FILE) >> $RESULT_FILE
+done
+# Now sort the file so that it is
```
LouisCAD, line 39: Looks like the sentence has been truncated.

alexvanyo: Nice catch, updated in 3bd4f53.

This was the case in quite a few places due to this file being moved around, so I don't think this will be on the only place to fix this 😄

AlwaysOnKotlin/Wearable/build.gradle
alexvanyo: Updated in the latest set of commits, with the context of https://issuetracker.google.com/issues/196420849

codingjeremy (approved):

PR #131, opened 2021-07-27
Watchface alpha rendering to latest alpha.

codingjeremy:


WatchFaceAlphaKotlin/app/build.gradle
```diff
@@ -66,13 +66,13 @@ dependencies {
     implementation 'androidx.wear:wear:1.1.0'
 
     // Watch face specific libraries
-    implementation "androidx.wear:wear-complications-data:$rootProject.watchFaceAlphaVersion"
-    implementation "androidx.wear:wear-complications-provider:$rootProject.watchFaceAlphaVersion"
+    implementation "androidx.wear:wear-complications-data:$rootProject.watchFaceVersion"
+    implementation "androidx.wear:wear-complications-provider:$rootProject.watchFaceProviderVersion"
```
alexvanyo, line 70: `wear-complications-provider` was replaced by `wear-complications-data-source`, so we should update that here.

I'm actually slightly surprised this works as is, I wonder if `wear-watchface-complications-rendering` transitively exposes `wear-complications-data-source`? And then no classes conflict?

WatchFaceAlphaKotlin/app/src/main/AndroidManifest.xml
```diff
@@ -78,7 +78,7 @@
         </service>
 
         <activity
-            android:name="androidx.wear.complications.ComplicationHelperActivity"
+            android:name="androidx.wear.watchface.ComplicationHelperActivity"
             android:exported="false" />
```
alexvanyo, line 82: I believe https://android-review.googlesource.com/c/platform/frameworks/support/+/1752949/ means that we can get rid of this declaration entirely, since it will automatically be added by the library's manifest.

codingjeremy: Sorry,  I forgot to merge this and I have a couple PRs that rely on this. I will fix in a follow up PR.

WatchFaceAlphaKotlin/app/build.gradle
codingjeremy: I was using alpha 17 where it didn't exist but, yeah, not sure how I got it.  :D

Oh well, I bumped to 19 and renamed it.

alexvanyo (approved):
Looks good!

PR #137, opened 2021-08-16
Add Tiles example to Readme

volkansahin45:

PR #138, opened 2021-08-17
Update AndroidX capitalizations

volkansahin45:
In my last PR I had copied the description from the original readme so I noticed its lowercase in here, too. #137


alexvanyo (approved):
Thanks!

PR #139, opened 2021-08-18
Adds support for user to edit watch face styles via Activity.

codingjeremy:


WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/AnalogWatchCanvasRenderer.kt
```diff
@@ -115,8 +115,14 @@ class AnalogWatchCanvasRenderer(
     private var currentWatchFaceSize = Rect(0, 0, 0, 0)
 
     init {
+
+        Log.d(TAG, "init { }")
+
         scope.launch {
             analogWatchFaceViewModel.uiState.collect { uiState ->
+
+                Log.d(TAG, "collect worked! : $uiState")
```
ADClarke, line 124: Might be useful to remove some of the logging once everything is working.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
```diff
@@ -1,17 +1,145 @@
...
+        val colorStyleIdAndResourceIdsList = enumValues<ColorStyleIdAndResourceIds>()
+        val numbers: IntRange = colorStyleIdAndResourceIdsList.indices
+
+        val random = numbers.random()
+        val newColorStyle = colorStyleIdAndResourceIdsList[random]
+        val newColorStyleId = newColorStyle.id
+
+        Log.d(TAG, "\tnewColorStyleId (random): $newColorStyleId")
+
+        val newColorStyleValue = ColorStyleIdAndResourceIds.toOption(
```
ADClarke, line 132: Could we instead try and get the right Option out of the UserStyleSchema rather than constructing a new one?

The EditorSession within WatchFaceConfigViewModel has a userStyleSchema property.  The UserStyleSchema is defined like so:

public class UserStyleSchema(
    public val userStyleSettings: List<UserStyleSetting>
)

It should be possible to find the color settings list from userStyleSettings (the editor should know the ID) on init. And then we can just iterate over the options list in the colorUserStyleSetting with findOrNull looking for a matching ID.

The advantage of this is we'd only need one place in the watch face code where the options are constructed, reducing the effort needed if a new color option was added.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigViewModel.kt
```diff
@@ -0,0 +1,265 @@
...
+            Log.d(WatchFaceConfigActivity.TAG, "userStyle: $userStyle")
+
+            extractsUserStyles(userStyle)
+            updatesWatchFacePreview()
+        }
+    }
+
+    private fun extractsUserStyles(newUserStyle: UserStyle) {
+        // Loops through user styles and retrieves user editable styles.
+        for (options: Map.Entry<UserStyleSetting, UserStyleSetting.Option> in newUserStyle.selectedOptions) {
```
ADClarke, line 129: Nit: I think you don't need to specify the type here unless you want to do so for clarity.

ADClarke (changes requested):
I'm not really up to speed with flows but generally looks good.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/WatchFaceUserSettingChangesDataSource.kt
```diff
@@ -37,7 +37,7 @@ class WatchFaceUserSettingChangesDataSource(
         val userStyleChangeListener = object : CurrentUserStyleRepository.UserStyleChangeListener {
             override fun onUserStyleChanged(userStyle: UserStyle) {
                 Log.d(TAG, "onUserStyleChanged(): userStyle: $userStyle")
-                trySend(userStyle)
+                trySend(UserStyle(userStyle))
```
alexvanyo, line 40: I believe this should no longer be necessary with `alpha20`, which included the changes to equals/hashcode for `UserStyle`.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
```diff
@@ -1,17 +1,145 @@
...
+        binding.minuteHandLengthSlider.addOnChangeListener { slider, value, fromUser ->
+            Log.d(TAG, "addOnChangeListener(): $slider, $value, $fromUser")
+            viewModel.setMinuteHandArmLength(value)
+        }
+
+        lifecycleScope.launch(Dispatchers.Main.immediate) {
+            viewModel = WatchFaceConfigViewModel(
+                    activity = this@WatchFaceConfigActivity,
+                    editIntent = intent
+                )
```
alexvanyo, line 68: We should never be constructing a AAC `ViewModel` directly, nor should we be giving it a reference to an `Activity` or `View`.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigViewModel.kt
```diff
@@ -0,0 +1,265 @@
...
+ * Handles all ops for saving changes from [WatchFaceConfigActivity] to the watch face data layer,
+ * that is to the [EditorSession].
+ *
+ * Uses the keys, [UserStyleSetting], for each of our user styles and sets their values
+ * [UserStyleSetting.Option]. After a new value is set, creates a new image preview via screenshot
+ * class and triggers a listener (which creates new data
+ * for the [StateFlow] that feeds back to the Activity).
+ */
+class WatchFaceConfigViewModel(activity: ComponentActivity, editIntent: Intent) : ViewModel() {
+    private val changeSupport = PropertyChangeSupport(this)
```
alexvanyo, line 62: The `PropertyChangeSupport` looks really suspicious. We should be able to use a `kotlinx.coroutines` primitive versus a callback/listener setup, and ideally even be able to replace the imperative "notify about an update" into a pure state flow.

LouisCAD:
@ADClarke Here's a good video on Fliws by one of its designers, Roman Elizarov, also Kotlin's current lead: https://youtu.be/tYcqn48SMT8

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/data/WatchFaceUserSettingChangesDataSource.kt
codingjeremy: Ha, good call, just saw alpha 20 was released today... I will revise.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
codingjeremy: Whoops, good catch, switched to property delegation.

I'm a little stuck on the other side. The EditorSession (basically the DB) needs an instance of ComponentActivity and Intent... I have a bug for it.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigViewModel.kt
codingjeremy: Will revise and follow up.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/AnalogWatchCanvasRenderer.kt
codingjeremy: I will do a pass after the things are addressed and strip out the Log.d()

WatchFaceAlphaKotlin/app/build.gradle
```diff
@@ -61,6 +65,14 @@ dependencies {
...
     api "org.jetbrains.kotlinx:kotlinx-coroutines-android:$rootProject.coroutines"
 
+    implementation "androidx.activity:activity-ktx:1.3.1"
+
+    // Lifecycle components
+    implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.3.1"
+    implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:2.3.1"
+
+    // Material components
+    api 'com.google.android.material:material:1.4.0'
```
alexvanyo, line 75: Small nit, but is there a reason that these are `api`? Unless another module is depending on this one, `api` doesn't really mean anything over `implementation`.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigViewModel.kt
```diff
@@ -0,0 +1,282 @@
...
+        userStyleSetting: UserStyleSetting,
+        userStyleOption: UserStyleSetting.Option
+    ) {
+        Log.d(WatchFaceConfigActivity.TAG, "setUserStyleOption()")
+        Log.d(WatchFaceConfigActivity.TAG, "\tuserStyleSetting: $userStyleSetting")
+        Log.d(WatchFaceConfigActivity.TAG, "\tuserStyleOption: $userStyleOption")
+
+        // Casts as a [HashMap], so we can update values.
+        val hashmap =
+            userStyle.selectedOptions as HashMap<UserStyleSetting, UserStyleSetting.Option>
```
alexvanyo, line 222: This cast looks unsafe and suspicious, I think there's a `toMutableUserStyle()` to avoid mutating one in-place.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
```diff
@@ -1,17 +1,147 @@
...
+    fun onClickColorStylePickerButton(view: View) {
+        Log.d(TAG, "onClickColorStylePickerButton() $view")
+
+        // TODO (codingjeremy): Replace with a RecyclerView to choose color style (next CL)
+        // Selects a random color style from list.
+        val colorStyleIdAndResourceIdsList = enumValues<ColorStyleIdAndResourceIds>()
+        val numbers: IntRange = colorStyleIdAndResourceIdsList.indices
+
+        val random = numbers.random()
+        val newColorStyle = colorStyleIdAndResourceIdsList[random]
```
alexvanyo, line 129: Small optimization here, we can just call `.random()` on the list directly I believe.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
```diff
@@ -1,17 +1,147 @@
...
 
-// TODO (codingjeremy): Add activity support to edit watch face (follow up PR).
+/**
+ * Allows user to edit certain parts of the watch face (color style, ticks displayed, minute arm
+ * length) by using the [WatchFaceConfigViewModel]. (All widgets are disabled until data is loaded.)
+ */
 class WatchFaceConfigActivity : ComponentActivity() {
+    private val viewModel: WatchFaceConfigViewModel by viewModels {
+        WatchFaceConfigViewModel.WatchFaceConfigViewModelFactory(
+            this@WatchFaceConfigActivity,
```
alexvanyo, line 40: Since the editor session requires an `Activity` (at least right now), we have to manage it within the `Activity`'s scope. If we pass it to the `ViewModel`, which will outlive the activity, we'll get memory leaks in the best case and in the worst case not work or crash since the editor is managing result contracts.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigViewModel.kt
codingjeremy: Ah, yes, that was just added in 20. I upgraded sample and swapped it.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
codingjeremy: Done

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigViewModel.kt
codingjeremy: Done; Removed callbackFlow and used a regular StateFlow.

WatchFaceAlphaKotlin/app/build.gradle
```diff
@@ -53,17 +57,25 @@ android {
...
 
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-    implementation 'androidx.core:core-ktx:1.6.0'
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$rootProject.kotlinVersion"
+    implementation "androidx.core:core-ktx:$rootProject.ktxVersion"
 
     // Kotlin components
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$rootProject.kotlinVersion"
     api "org.jetbrains.kotlinx:kotlinx-coroutines-core:$rootProject.coroutines"
```
alexvanyo, line 65: Nit: I know it wasn't touched by this PR, but it'd be good to switch these to `implementation` too

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/AnalogWatchCanvasRenderer.kt
```diff
@@ -203,6 +210,7 @@ class AnalogWatchCanvasRenderer(
     }
 
     override fun onDestroy() {
+        Log.d(TAG, "onDestroy()")
```
alexvanyo, line 213: Nit: did you want to remove this one too?

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
```diff
@@ -1,17 +1,137 @@
...
+        Log.d(TAG, "\tselected color style: $colorStyleId")
+
+        binding.ticksEnabledSwitch.isChecked = userStylesAndPreview.ticksEnabled
+        binding.minuteHandLengthSlider.value = userStylesAndPreview.minuteHandLength
+        binding.preview.watchFaceBackground.setImageBitmap(userStylesAndPreview.previewImage)
+
+        if (!initialDataLoaded) {
+            initialDataLoaded = true
+            enabledWidgets()
+        }
```
alexvanyo, line 104: Nit: just a small personal preference, but repeatedly setting `isEnabled=true` here doesn't have any downside, so why not just always set them? We avoid needing a flag, and makes our state handling overall more idempotent.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
```diff
@@ -1,17 +1,137 @@
...
+        binding.colorStylePickerButton.isEnabled = true
+        binding.ticksEnabledSwitch.isEnabled = true
+        binding.minuteHandLengthSlider.isEnabled = true
+    }
+
+    override fun onDestroy() {
+        Log.d(TAG, "onDestroy()")
+        stateHolder.onCleared()
+        // Makes sure the activity closes.
+        finish()
```
alexvanyo, line 117: Just curious, why do we need this here?

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigStateHolder.kt
```diff
@@ -0,0 +1,245 @@
...
+        userStyleOption: UserStyleSetting.Option
+    ) {
+        Log.d(WatchFaceConfigActivity.TAG, "setUserStyleOption()")
+        Log.d(WatchFaceConfigActivity.TAG, "\tuserStyleSetting: $userStyleSetting")
+        Log.d(WatchFaceConfigActivity.TAG, "\tuserStyleOption: $userStyleOption")
+
+        val mutableUserStyle = editorSession.userStyle.toMutableUserStyle()
+        mutableUserStyle[userStyleSetting] = userStyleOption
+        editorSession.userStyle = mutableUserStyle.toUserStyle()
+        updatesWatchFacePreview()
```
alexvanyo, line 206: There's a race here, since the initialization of `editorSession` occurs asynchronously. It's possible that a user updates an option before `editorSession` is initialized, which would cause a crash.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigStateHolder.kt
```diff
@@ -0,0 +1,245 @@
...
+    }
+
+    fun onCleared() {
+        editorSession.close()
```
alexvanyo, line 210: Ditto here for a race, although this one is a bit easier to take care of: we can check if the `editorSession` is initialized before trying to close it.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigStateHolder.kt
alexvanyo: Ah, I remembered that the buttons will only be enabled once the editor session is initialized. So we should be good here then.

WatchFaceAlphaKotlin/app/build.gradle
codingjeremy: Done

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/AnalogWatchCanvasRenderer.kt
codingjeremy: Removed

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
codingjeremy: I'm confused, it's only set to true once and never done again, i.e., it's false by default and so this block will only be executed when it is false. After that, it sets it to true, enables the widgets and is never called again.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
codingjeremy: Good point, I need to add a comment.

It's called out specifically to do in the [jetpack code](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/wear-watchface-editor/samples/src/main/java/androidx/wear/watchface/editor/sample/WatchFaceConfigActivity.kt;l=216?q=EditorSession&ss=androidx%2Fplatform%2Fframeworks%2Fsupport). I will start a thread with Alex and verify why... then add a follow up CL with a comment explaining why.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigStateHolder.kt
codingjeremy: I added a note at the top of the class to call that out.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigStateHolder.kt
codingjeremy: Added check.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
alexvanyo: Right, I was think we could just do this:

```suggestion
        enabledWidgets()
```

Doesn't change any behavior, but reduces the number of branches to worry about.

alexvanyo (changes requested):
It looks like the CI is failing due to a `RestrictedApi` lint error

PR #141, opened 2021-08-18
[RuntimePermissions] Modernize phone app

alexvanyo:
Updates the phone portion of the `RuntimePermissions` app, converting from Java to Kotlin, introducing coroutines, and other best practices.

I also switched the phone permission from storage to the cellular information, since the old behavior (showing all of the folders in the emulated storage) is deprecated.


RuntimePermissionsWear/Application/build.gradle
```diff
@@ -20,17 +20,23 @@ plugins {
 }
 
 android {
-    compileSdkVersion 30
+    compileSdkVersion 31
```
codingjeremy, line 23: Is this just so it will build right now?

codingjeremy (approved):
Just some minor nits.

RuntimePermissionsWear/Application/src/main/java/com/example/android/wearable/runtimepermissions/IncomingRequestPhoneService.kt
```diff
@@ -0,0 +1,158 @@
...
+            }
+        }
+    }
+
+    private suspend fun promptUserForPhonePermission(nodeId: String) {
+        val phoneInfoPermissionApproved =
+            (
+                ActivityCompat.checkSelfPermission(this, phoneSummaryPermission)
+                    == PackageManager.PERMISSION_GRANTED
+                )
```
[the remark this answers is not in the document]
alexvanyo: Oh yeah, this looks weird... it looks like there's an unnecessary set of parentheses around the condition. I'll remove that.

RuntimePermissionsWear/Application/build.gradle
alexvanyo: Ah, here I'm updating the phone app. So no reason to avoid going to API 31!

alexvanyo:
Thanks @codingjeremy !

PR #150, opened 2021-09-01
[TimeText] Move to standalone sample

alexvanyo:
This PR removes the `timetext` library from the `Speaker` sample, and turns it into a standalone sample.

I add a super simple showcase sample highlighting the usage, as well as a README with usage documentation.


alexvanyo:
Closed in favor of #151

PR #152, opened 2021-09-04
Adds edit support for the complications.

codingjeremy:
Adds support for editing the complications and fixes any rendering issues with changes to the complications.

Also, does some slight cleanup.

### DIFF
--- a/WatchFaceAlphaKotlin/app/src/main/AndroidManifest.xml
+++ b/WatchFaceAlphaKotlin/app/src/main/AndroidManifest.xml
@@ -49,13 +49,9 @@
             android:permission="android.permission.BIND_WALLPAPER">
             <intent-filter>
                 <action android:name="android.service.wallpaper.WallpaperService" />
-
                 <category android:name="com.google.android.wearable.watchface.category.WATCH_FACE" />
             </intent-filter>
 
-            <meta-data
-                android:name="com.google.android.wearable.standalone"
-                android:value="true" />
             <meta-data
                 android:name="com.google.android.wearable.watchface.preview"
                 android:resource="@drawable/watch_preview" />
@@ -63,11 +59,15 @@
                 android:name="com.google.android.wearable.watchface.preview_circular"
                 android:resource="@drawable/watch_preview" />
             <meta-data
-                android:name="com.google.android.wearable.watchface.wearableConfigurationAction"
-                android:value="androidx.wear.watchface.editor.action.WATCH_FACE_EDITOR" />
-            <meta-data
                 android:name="android.service.wallpaper"
                 android:resource="@xml/watch_face" />
+            <meta-data
+                android:name="com.google.android.wearable.watchface.wearableConfigurationAction"
+                android:value="androidx.wear.watchface.editor.action.WATCH_FACE_EDITOR" />
+
+            <meta-data
+                android:name="com.google.android.wearable.watchface.companionBuiltinConfigurationEnabled"
+                android:value="true" />
         </service>
 
         <activity
--- a/WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
+++ b/WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
@@ -86,6 +86,11 @@ class WatchFaceConfigActivity : ComponentActivity() {
         }
     }
 
+    override fun onDestroy() {
+        stateHolder.onCleared()
+        super.onDestroy()
+    }
+
     private fun updateWatchFacePreview(
         userStylesAndPreview: WatchFaceConfigStateHolder.UserStylesAndPreview
     ) {
--- a/WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
+++ b/WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
@@ -25,6 +25,8 @@ import com.example.android.wearable.alpha.databinding.ActivityWatchFaceConfigBin
 import com.example.android.wearable.alpha.editor.WatchFaceConfigStateHolder.Companion.MINUTE_HAND_LENGTH_DEFAULT_FOR_SLIDER
 import com.example.android.wearable.alpha.editor.WatchFaceConfigStateHolder.Companion.MINUTE_HAND_LENGTH_MAXIMUM_FOR_SLIDER
 import com.example.android.wearable.alpha.editor.WatchFaceConfigStateHolder.Companion.MINUTE_HAND_LENGTH_MINIMUM_FOR_SLIDER
+import com.example.android.wearable.alpha.utils.LEFT_COMPLICATION_ID
+import com.example.android.wearable.alpha.utils.RIGHT_COMPLICATION_ID
 import kotlinx.coroutines.Dispatchers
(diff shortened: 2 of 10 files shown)


WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
```diff
@@ -107,8 +109,6 @@ class WatchFaceConfigActivity : ComponentActivity() {
 
     override fun onDestroy() {
         stateHolder.onCleared()
-        // Makes sure the activity closes.
-        finish()
```
alexvanyo, line 111: 👍

WatchFaceAlphaKotlin/app/src/main/res/layout/watch_face_and_complications_preview.xml
```diff
@@ -16,21 +16,51 @@
...
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="@dimen/analog_complication_settings_preview_size"
+    android:layout_height="@dimen/analog_complication_settings_preview_size"
+    tools:context=".editor.WatchFaceConfigActivity">
+
+    <Button
+        android:id="@+id/left_complication"
+        android:layout_width="0dp"
+        android:layout_height="match_parent"
```
alexvanyo, line 31: Small nit: we should avoid `match_parent` in `ConstraintLayout` children, and instead use `0dp`:

https://developer.android.com/training/constraint-layout#adjust-the-view-size

WatchFaceAlphaKotlin/app/src/main/AndroidManifest.xml
```diff
@@ -49,25 +49,25 @@
             android:permission="android.permission.BIND_WALLPAPER">
             <intent-filter>
                 <action android:name="android.service.wallpaper.WallpaperService" />
-
                 <category android:name="com.google.android.wearable.watchface.category.WATCH_FACE" />
             </intent-filter>
 
-            <meta-data
-                android:name="com.google.android.wearable.standalone"
-                android:value="true" />
```
alexvanyo, line 58: Is this tag no longer required/included automatically?

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
```diff
@@ -123,6 +123,16 @@ class WatchFaceConfigActivity : ComponentActivity() {
         stateHolder.setColorStyle(newColorStyle.id)
     }
 
+    fun onClickLeftComplicationButton(view: View) {
+        Log.d(TAG, "onClickLeftComplicationButton() $view")
+        stateHolder.setComplication(LEFT_COMPLICATION_ID)
+    }
+
+    fun onClickRightComplicationButton(view: View) {
+        Log.d(TAG, "onClickLeftComplicationButton() $view")
```
alexvanyo, line 132: Small nit (assuming these logs will be removed once the issue is cleared up), but this one also says `onClickLeft` instead of `onClickRight`

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigStateHolder.kt
```diff
@@ -212,6 +246,7 @@ class WatchFaceConfigStateHolder(
         if (::editorSession.isInitialized) {
             editorSession.close()
         }
+        scope.cancel()
```
alexvanyo, line 249: It feels a bit weird for `scope.cancel()` to be called here, since the scope is being passed in from above. If something else is in charge of creating `scope`, I would expect that thing to also manage cancelling it.

WatchFaceAlphaKotlin/app/src/main/res/layout/watch_face_and_complications_preview.xml
codingjeremy: Good call, I replaced all of them.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/utils/ComplicationUtils.kt
```diff
@@ -80,20 +80,18 @@ fun createComplicationSlotManager(
     drawableId: Int = DEFAULT_COMPLICATION_STYLE_DRAWABLE_ID
 ): ComplicationSlotsManager {
 
-    val defaultComplicationDrawable = ComplicationDrawable.getDrawable(context, drawableId)
-
-    val canvasComplicationFactory =
+    val canvasLeftComplicationFactory =
```
alexvanyo, line 83: I think just removing the `defaultComplicationDrawable` should be enough here, and we can use the same `CanvasComplicationFactory` in both places.

codingjeremy: Moved it down to be separately called for each Factory and that works.

WatchFaceAlphaKotlin/app/src/main/AndroidManifest.xml
codingjeremy: It still exists at the top level, somehow one ended up in the service itself.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
codingjeremy: Fixed.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigStateHolder.kt
codingjeremy: You are right, it is a lifecycleScope... removed.

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
```diff
@@ -105,13 +107,6 @@ class WatchFaceConfigActivity : ComponentActivity() {
         binding.minuteHandLengthSlider.isEnabled = true
     }
 
-    override fun onDestroy() {
-        stateHolder.onCleared()
```
alexvanyo, line 109: Ah, we still need to call `onCleared` here, right? Otherwise nothing will call that for the `StateHolder`.
There might be a more composable version here where we give the state holder a `Lifecycle` reference to clear itself up.

alexvanyo (commented):
Looks good, just one more nit to add back the clearing of the `StateHolder` (for things separate from the `CoroutineScope`)

WatchFaceAlphaKotlin/app/src/main/java/com/example/android/wearable/alpha/editor/WatchFaceConfigActivity.kt
codingjeremy: oh geez, sorry, thanks, good catch, I need that to close the editor session.

alexvanyo (approved):
LGTM!

codingjeremy:
Thanks!

PR #154, opened 2021-09-08
[Speaker] Convert to Compose

alexvanyo:
Converts the Speaker sample to Jetpack Compose.

This is mostly an in-place conversion, adapting the logic and components to use Compose for Wear OS.


WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        SpeakerScreen(
+            appState = stateHolder.appState,
+            isPermissionDenied = stateHolder.isPermissionDenied,
+            recordingProgress = stateHolder.recordingProgress,
+            onMicClicked = {
+                coroutineScope.launch {
+                    val job = launch { stateHolder.onAction(AppAction.MicClicked) }
+                    lifecycleOwner.lifecycle.awaitEvent(Lifecycle.Event.ON_STOP)
```
yschimke, line 106: I'm curious about this pattern. onAction doesn't look like a suspending function, does the cancellation interrupt the onAction method?

Also wondering why launching with coroutineScope instead of either LaunchedEffect or using LocalLifecycleOwner.current.lifecycleScope which I assume would handle the cancellation?

I'm relatively new to Compose, so mainly trying to understand why the simpler forms don't work here.  Thanks for your patience :)

alexvanyo: > I'm relatively new to Compose
So am I, and aren't we all 😄 

`onAction` is a suspending function, which will modify the underlying state, as well as asynchronously do things in response to the user action. In particular, this allows tying the cancellation of `onAction` to cancelling recording and playback, since those are encapsulated as suspending functions as well.

So in this case (assuming the user has granted permission), `onAction(MicClicked)` will kick off the recording, updating the overall state to be `recording`, updating the progress for the duration of the recording, and resetting the layout when recording is finished.

Since we're doing all of that in response to a user action, I'm launching into a `rememberCoroutineScope()` (as opposed to a `LaunchedEffect` or similar, which I would do if it was in response to something other than a user action).

There are two other behaviors that also complicate this:
- Inside the state holder is a [`MutatorMutex`](https://developer.android.com/reference/kotlin/androidx/compose/foundation/MutatorMutex), provided by Compose. This allows cancelling other coroutines, so we get a lot of the state management for free: If the user clicks record, and then cancels it, the `onAction` for recording will be automatically cancelled, and cleaned up.
- I want recording and playback to be cancelled upon the app going to the background (or in other words, I only want it to occur when at least started). So the `lifecycleScope` would allow for nicely cleaning everything up once destroyed, but not as nicely for stopped.

alexvanyo: Thinking about this more, there's one thing my current approach doesn't do:

If the `onAction` completes successfully, we don't really have to keep waiting for the `ON_STOP` even. Really what we want to do is ["race"](https://github.com/Kotlin/kotlinx.coroutines/issues/2867) waiting for `ON_STOP` and the `onAction` call.

Whichever happens first, we want to cancel the other.

yschimke: What would break if you changed

```
            onMicClicked = {
                scope.launch {
                    raceOf(
                        { stateHolder.onAction(AppAction.MicClicked) },
                        { lifecycleOwner.lifecycle.awaitEvent(Lifecycle.Event.ON_STOP) }
                    )
                }
            },
```

to

```
            onMicClicked = {
                scope.launch {
                     stateHolder.onAction(AppAction.MicClicked)
                }
            },
```            

Does rememberCoroutineScope get closed when the overall Activity is stopped, or the @Composable is removed form the UI?

This is the sort of nuanced code that seems difficult to get right. As a sample I hope (but don't know if possible) that there is some simple obvious compose idiomatic way where this just works without everyone writing similarly complex code.

yschimke: I wonder whether lifecycle-runtime-ktx is missing something here.  It feels like what you are trying is directly in scope.

https://developer.android.com/topic/libraries/architecture/lifecycle
> Pausing and resuming animated drawables. Use lifecycle-aware components to handle pausing animated drawables when while app is in the background and resume drawables after the app is in the foreground.

yschimke: I wonder whether `withStateAtLeast` could be the solution here? As a sample, I'm concerned if you need to understand this complicated state dance, others will to.

alexvanyo: Yeah, `withStateAtLeast` almost does what I want, but the `block` passed to it can't be suspending.

If there was a version of `withStateAtLeast` that did take a suspending block, that'd roughly be the behavior I'd expect:
wait for the state to be met, start the block, and cancel it if dropping below the given state.

yschimke: Thanks for humoring me with these questions...

WearSpeakerSample/wear/build.gradle
```diff
@@ -45,19 +45,30 @@ android {
...
     implementation "androidx.core:core-ktx:1.6.0"
-    implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.4.0-alpha03"
-    implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:2.4.0-alpha03"
-    implementation "androidx.media:media:1.4.1"
-    implementation "androidx.percentlayout:percentlayout:1.0.0"
-    implementation "androidx.vectordrawable:vectordrawable-animated:1.1.0"
-    implementation "androidx.wear:wear:1.1.0"
+    implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.4.0-beta01"
+    implementation "androidx.media:media:1.4.2"
+    implementation "androidx.wear.compose:compose-foundation:1.0.0-alpha06"
```
JolandaVerhoef, line 71: You can remove this dependency

WearSpeakerSample/wear/build.gradle
```diff
@@ -45,19 +45,30 @@ android {
...
-    implementation "androidx.constraintlayout:constraintlayout:2.1.0"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.2"
+    implementation "androidx.activity:activity-compose:1.3.1"
+    // TODO: Remove direct dependency on androidx.compose.material:material, and only depend on
+    //       androidx.wear.compose:compose-material
+    //       This is only used for LinearProgressIndicator
+    implementation 'androidx.compose.material:material:1.0.2'
+    implementation 'androidx.compose.material:material-ripple:1.0.2'
+    implementation 'androidx.compose.material:material-icons-extended:1.0.2'
+    implementation 'androidx.compose.ui:ui:1.0.2'
```
JolandaVerhoef, line 65: You can remove this dependency

WearSpeakerSample/wear/build.gradle
```diff
@@ -45,19 +45,30 @@ android {
...
+    implementation 'androidx.compose.ui:ui-tooling:1.0.2'
+    implementation "androidx.constraintlayout:constraintlayout-compose:1.0.0-beta02"
     implementation "androidx.core:core-ktx:1.6.0"
```
JolandaVerhoef, line 68: You can remove this dependency

JolandaVerhoef (commented):
Have you considered using Navigation in this sample? If I look at the app, it seems logical to create separate screens for the dashboard, and each of the three functionalities. You could then use the Accompanist Navigation library to transition between them: https://google.github.io/accompanist/navigation-animation/

JolandaVerhoef:
Can you explain how I can get a dialog to show? I'm trying to get the rationale dialog to show, but no matter which answer I give on the permission dialog, it doesn't show.. And I don't have an emulator without a speaker.

JolandaVerhoef:
I'm not a big fan of this conversion. This sample should focus on the speaker capabilities of a watch. Instead, it now contains a lot of complex code to deal with Compose. 

This leads to a higher maintenance burden, which might not be what we want.

If we can greatly simplify this sample's UI, I think writing it in Compose still makes sense. 

Would it be an option to change it to:
- A basic list with three items: "Record", "Play recording", and "Play music"
- Navigation to go from one to the next, and back with the back gesture.
- View model scoping while recording or doing playback, which closes when the user closes the screen or app.

WearSpeakerSample/README.md
```diff
@@ -1,11 +1,14 @@
...
-Android WearSpeakerSample Sample
+WearSpeakerSample
 ================================
 
 A sample that shows how you can record voice using the microphone on a wearable and
 play the recorded voice or an mp3 file, if the wearable device is connected to a speaker
 (bluetooth or built-in).
 
+This sample is also written entirely with Jetpack Compose, using [Wear-specific components provided
+by AndroidX](https://developer.android.com/jetpack/androidx/releases/wear-compose).
```
codingjeremy, line 10: We should call out this is an advanced Compose sample and why that is so devs who find themselves reading this README really understand that.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -16,161 +16,212 @@
...
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
+import java.time.Duration
+import kotlin.coroutines.resume
 
 /**
- * A state holder driving the logic of [MainActivity].
+ * A state holder driving the logic of the app.
```
codingjeremy, line 39: Can we call out why application and activity are needed here, i.e., activity is required to record audio, permissions, etc.

codingjeremy (commented):
Being newer to Compose, take my opinion with a grain of salt. ^_^;;

I agree that, at least for me, it was really hard for me to understand this sample based on its complexity. I had to re-read things many times to halfway grok what was going on... and I'm not 100% sure I still get it. 

As Jolanda pointed out, can we simplify this with Navigation? 

I really like the UI, so I would like to keep that vs. a list, but maybe if this sample is simplified with Navigation, we can still use the UI?

It could be one of our more advanced samples, but we can wait to submit it until we have a simple sample and a list/detail sample. (We can clearly call that out in the top-level README.)

Again, it could be I just need to read it more times to "get it", but I don't want to scare away devs.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/ControlDashboard.kt
```diff
@@ -0,0 +1,242 @@
...
+ * The state for this class is driven by a [ControlDashboardState], which contains a [ControlDashboardButtonState]
+ * for each of the three buttons.
+ */
+@Composable
+fun ControlDashboard(
+    controlDashboardState: ControlDashboardState,
+    onMicClicked: () -> Unit,
+    onPlayClicked: () -> Unit,
+    onMusicClicked: () -> Unit,
+    modifier: Modifier = Modifier
```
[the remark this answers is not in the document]
yschimke: Not sure about ordering, but I'd say the key is named parameters and ant obvious content lambda last. Almost all uses wi be with named parameters.

WearSpeakerSample/README.md
yschimke: I think what you've described is a call to ask the compose core team how to implement this without that extra non obvious code. If its something missing in compose maybe we can get a few methods added or a different way to achieve it.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/ControlDashboard.kt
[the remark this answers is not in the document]
alexvanyo: This is the primary guidance:

https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#elements-accept-and-respect-a-modifier-parameter

In particular, for elements that intrinsically take up space, there should always be a `modifier` parameter that's the first optional parameter.

I'm not sure if there's an explicit guidance for state vs events, but state first seems to make sense to me.

alexvanyo:
I realized everything was already in place for adding animations back, without any need for full `MotionLayout` support. Since we can define everything by state anyway, the last commit to do so was extremely simple.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/ControlDashboard.kt
```diff
@@ -0,0 +1,268 @@
...
+    init {
+        // Check that at most one of the buttons is expanded
+        require(
+            listOf(
+                micState.expanded,
+                playState.expanded,
+                musicState.expanded
+            ).map {
+                if (it) 1 else 0
+            }.sum() <= 1
```
JolandaVerhoef, line 265: replace with `.count { it } <= 1`

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -16,161 +16,215 @@
...
+ *
+ * This state holder is scoped to the composition (and therefore can be provided an [Activity]) to handle
+ * permission-related logic.
```
JolandaVerhoef, line 42: @manuelvicnt Can you double check that this is the right approach?

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -16,161 +16,215 @@
...
-
     /**
-     * The backing [MutableStateFlow] for the [appState].
+     * The [MutatorMutex] that guards the state of the app.
+     *
+     * Due to being a [MutatorMutex], this automatically handles cleanup of any ongoing asynchronous work,
+     * like playing music or recording, ensuring that only one operation is occurring at a time.
      */
-    private val appStateMutableStateFlow = MutableStateFlow<AppState>(AppState.Ready(transitionInstantly = true))
+    private val appStateMutatorMutex = MutatorMutex()
```
JolandaVerhoef, line 56: @manuelvicnt @JoseAlcerreca I'd like to get your input on this AppState implementation.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * A helper method to run the given [block] when the state is at least [state].
+ *
+ * This will wait until the given [state] is achieved, and then will run the block.
+ * If the state drops back down while the [block] is running, the [block] will be cancelled.
+ *
+ * This method resumes successfully when the [block] finishes (either normally, or after being cancelled)
+ */
+private suspend fun Lifecycle.withStateAtLeast(state: Lifecycle.State, block: suspend () -> Unit) {
```
JolandaVerhoef, line 141: Would you be able to use the Lifecycle's own method here? https://developer.android.com/reference/androidx/lifecycle/Lifecycle.State#isAtLeast(androidx.lifecycle.Lifecycle.State)

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/ControlDashboard.kt
alexvanyo: Ah, that's way cleaner!

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
alexvanyo: Unfortunately no, since `isAtLeast` is a one-time check.

We could use `isAtLeast` to ensure that the state is started when first launching the block, but we wouldn't get any of the cancellation behavior if the state then drops below started.

WearSpeakerSample/gradle.properties
```diff
@@ -10,7 +10,7 @@
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx10248m -XX:MaxPermSize=256m
-# org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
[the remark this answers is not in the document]
alexvanyo: That's a good question... I think I added these to add more memory for CI, but I just tried removing them and CI still passed. Removed in latest commits.

alexvanyo:
@JoseAlcerreca and @manuelvicnt: Hi, just added you as reviewers for the state holder implementation. Please let me know what you think!

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/ControlDashboard.kt
```diff
@@ -0,0 +1,266 @@
...
+                .fillMaxSize()
+                .padding(iconPadding)
+        )
+    }
+}
+
+/**
+ * The state for a single [ControlDashboardButton].
+ */
+data class ControlDashboardButtonState(
```
manuelvicnt, line 241: Should this be called `ControlDashboardButtonUiState`? This class is a data class that doesn't handle any UI logic and whose state cannot be mutated. Following our naming conventions, it should be `XUiState`

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/ControlDashboard.kt
```diff
@@ -0,0 +1,266 @@
...
+    val expanded: Boolean,
+    val enabled: Boolean,
+    val visible: Boolean
+)
+
+/**
+ * The state for a [ControlDashboard].
+ */
+data class ControlDashboardState(
```
manuelvicnt, line 250: Ditto, this should be `ControlDashboardUiState`

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -16,161 +16,215 @@
...
+import kotlin.coroutines.resume
 
 /**
- * A state holder driving the logic of [MainActivity].
+ * A state holder driving the logic of the app.
+ *
+ * This state holder is scoped to the composition (and therefore can be provided an [Activity]) to handle
+ * permission-related logic.
  */
 class MainStateHolder(
```
manuelvicnt, line 44: This class should be called `MainState` to be consistent with Compose's naming convention

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -16,161 +16,215 @@
...
-    val appState = appStateMutableStateFlow.asStateFlow()
+    var appState by mutableStateOf<AppState>(AppState.Ready(transitionInstantly = true))
+        private set
 
     /**
-     * The backing [MutableStateFlow] for the [isPermissionDenied] flow.
+     * The progress of an ongoing recording.
      */
-    private val isPermissionDeniedMutableStateFlow = MutableStateFlow(false)
+    var recordingProgress by mutableStateOf(0f)
```
manuelvicnt, line 67: This looks like it could be added to the `AppState.Recording` class. Is there a reason why we're not doing that? If so, document it as it might be confusing

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -16,161 +16,215 @@
...
+        private set
 
     /**
-     * A [StateFlow] of whether we know the user has denied the record audio permission.
+     * `true` if we know the user has denied the record audio permission.
      */
-    val isPermissionDenied = isPermissionDeniedMutableStateFlow.asStateFlow()
+    var isPermissionDenied by mutableStateOf(false)
```
manuelvicnt, line 73: Let's explain why this is not part of `AppState`

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
manuelvicnt: The usage of the `MutatorMutex` looks ok but I wonder if it would be simpler to just have the `appState` backed by a `StateFlow` and use the `update` function that gives you atomic updates out of the box.

I'm not sure if this will be the way of atomically setting values for `mutableStateOf`s in the future as you have also transactions in Compose.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -16,161 +16,215 @@
...
-                    ContextCompat.checkSelfPermission(application, Manifest.permission.RECORD_AUDIO) ==
-                    PackageManager.PERMISSION_GRANTED
-                ) {
-                    // The user granted the permission, continue on to start recording
-                    AppState.Recording
-                } else {
-                    // We have confirmation now that the user denied the permission
-                    isPermissionDeniedMutableStateFlow.value = true
-                    AppState.Ready(transitionInstantly = false)
+                AppAction.PermissionResultReturned ->
```
manuelvicnt, line 159: For this version, maybe it's ok to handle permissions in this way. But we should iterate on this implementation and make it more Compose friendly, using the new activityForResult APIs or the Accompanist permission APIs.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -16,161 +16,215 @@
...
+
+    /**
+     * A helper function to record, updating the progress state while recording.
+     *
+     * This requires the [Manifest.permission.RECORD_AUDIO] permission to run.
+     */
+    @RequiresPermission(Manifest.permission.RECORD_AUDIO)
+    private suspend fun recordUpdatingState() {
+        // We have the permission, we can start recording now
+        appState = AppState.Recording
```
manuelvicnt, line 184: In these two last functions, the mutex isn't used. Use it or reconsider using `StateFlow`

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -16,161 +16,215 @@
...
+            recordingJob.cancel()
+        }
+
+        appState = AppState.Ready(transitionInstantly = false)
+    }
+
+    /**
+     * Plays the MP3 file embedded in the application, updating the state to reflect the playing.
+     */
+    private suspend fun playMusicUpdatingState() {
```
manuelvicnt, line 210: `playMusicUpdatingState` doesn't seem like a good method name. Maybe simply have `playMusic` and defer the updating state logic to the caller of the function?

manuelvicnt: Ditto for `recordUpdatingState`

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
```diff
@@ -180,12 +234,45 @@ class MainStateHolder(
...
+    data class Ready(
+        /**
+         * If true, don't animate to this position
+         */
+        val transitionInstantly: Boolean,
+    ) : AppState()
+
+    object PlayingVoice : AppState()
+    object PlayingMusic : AppState()
+    object Recording : AppState()
```
manuelvicnt, line 277: As mentioned above, the `recordingProgress` state could be here to centralize all the info

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
```diff
@@ -0,0 +1,223 @@
...
+                }
+            )
+        }
+
+        val lifecycleOwner = LocalLifecycleOwner.current
+
+        requestPermissionLauncher = rememberLauncherForActivityResult(RequestPermission()) {
+            // We ignore the direct result here, since we're going to check anyway.
+            scope.launch {
+                lifecycleOwner.lifecycle.withStateAtLeast(Lifecycle.State.STARTED) {
```
manuelvicnt, line 91: Why do you need to wait for the Lifecycle to be at least started to update the state holder?

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
```diff
@@ -0,0 +1,223 @@
...
+                    stateHolder.onAction(AppAction.PermissionResultReturned)
+                }
+            }
+        }
+
+        // Notify the state holder whenever we become started to reset the state
+        LaunchedEffect(stateHolder) {
+            lifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
+                stateHolder.onAction(AppAction.Started)
```
manuelvicnt, line 100: Why this is needed? Looks like a code smell :(

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
```diff
@@ -0,0 +1,223 @@
...
+            }
+        }
+
+        SpeakerScreen(
+            appState = stateHolder.appState,
+            isPermissionDenied = stateHolder.isPermissionDenied,
+            recordingProgress = stateHolder.recordingProgress,
+            onMicClicked = {
+                scope.launch {
+                    lifecycleOwner.lifecycle.withStateAtLeast(Lifecycle.State.STARTED) {
```
manuelvicnt, line 110: These `withStateAtLeast` calls seem unnecessary since when buttons are clicked the app is supposed to be STARTED. Even if the `onAction` hasn't been processed. You might not want to delay and the app moves to the background, you might not want to delay this until the app goes back to the foreground

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
manuelvicnt: Added comments above, I don't think these checks are necessary. I need to be convinced :D

manuelvicnt: This method and the ones below (i.e. `raceOf` and `awaitEvents`) seem to add unnecessary complexity to this sample

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerScreen.kt
```diff
@@ -0,0 +1,173 @@
...
+) {
+    Scaffold(
+        modifier = Modifier.background(colorResource(id = R.color.black)),
+        timeText = {
+            TimeText()
+        }
+    ) {
+        // Determine the control dashboard state.
+        // This converts the main app state into a control dashboard state for rendering
+        val controlDashboardState by derivedStateOf {
```
manuelvicnt, line 66: Extract this out to a function in this file? Adds a bit of noise and doesn't actually help me understand what `SpeakerScreen` is doing

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerScreen.kt
```diff
@@ -0,0 +1,173 @@
...
+            ) {
+                LinearProgressIndicator(
+                    progress = recordingProgress
+                )
+            }
+        }
+    }
+}
+
+class AppStatePreviewProvider : CollectionPreviewParameterProvider<AppState>(
```
manuelvicnt, line 145: private?

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
alexvanyo: The `MutatorMutex` here is doing more work than just atomically updating `mutableStateOf`:

Since I'm expressing all of the recording and playback as `suspend` functions, I'm using the `MutatorMutex` to ensure that only one thing is happening at a time.

So if the user taps on the "Play music" button, and then taps on it again halfway through the song, the second tap will interrupt the playback due to the `appStateMutatorMutex. mutate` call, and the old playback will be cleaned up using the normal coroutine cancellation mechanisms.

An alternative solution here would be keeping track of an ongoing `Job` directly, but that solution ends up looking like a `MutatorMutex` anyway.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
alexvanyo: Let me know what you think of the latest approach. Instead of using the `withStateAtLeast` extension, I switched this out to send off an `onStopped` event to the state holder.

That will more cleanly cancel any ongoing playback, and reset the state to ready without the need for the custom extension.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
alexvanyo: Let me know what you think of the latest changes. I removed these custom `withStateAtLeast` checks, in favor of a more direct `onStopped` callback for the state holder, which will cancel ongoing work.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
alexvanyo: I added a comment to explain the reasoning!

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
alexvanyo: I am using the `rememberLauncherForActivityResult` API here, and passing it into the state holder. I couldn't use the Accompanist permission API here, since I want to immediately do something upon the user actively accepting the permission, so I want a callback for when the permission result is returned.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
alexvanyo: I renamed `AppState` to `PlaybackState`, since the former implies that all of the state is contained in it, instead of having it separated out as appropriate.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
alexvanyo: Done in the latest commit!

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerScreen.kt
```diff
@@ -142,14 +107,45 @@ fun SpeakerScreen(
     }
 }
 
-class AppStatePreviewProvider : CollectionPreviewParameterProvider<AppState>(
+private fun createControlDashboardUiState(
```
manuelvicnt, line 110: This should be remembered, otherwise it'll be created on every recomposition

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
```diff
@@ -88,130 +81,46 @@ fun SpeakerApp() {
...
-                    stateHolder.onAction(AppAction.PermissionResultReturned)
-                }
+                mainState.permissionResultReturned()
             }
         }
 
-        // Notify the state holder whenever we become started to reset the state
-        LaunchedEffect(stateHolder) {
+        // Notify the state holder whenever we become stopped to reset the state
+        LaunchedEffect(mainState, scope) {
```
manuelvicnt, line 89: Wouldn't be better to add a callback to the `LocalLifecycleOwner`? I think it's easier to understand

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainState.kt
```diff
@@ -0,0 +1,274 @@
...
+    private val showPermissionRationale: () -> Unit,
+    private val showSpeakerNotSupported: () -> Unit
+) {
+    /**
+     * The [MutatorMutex] that guards the state of the app.
+     *
+     * Due to being a [MutatorMutex], this automatically handles cleanup of any ongoing asynchronous work,
+     * like playing music or recording, ensuring that only one operation is occurring at a time.
+     */
+    private val appStateMutatorMutex = MutatorMutex()
```
manuelvicnt, line 56: Nit: `appStateMutex`? MutatorMutex is an impl detail

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainState.kt
```diff
@@ -0,0 +1,274 @@
...
+    private val soundRecorder = SoundRecorder(activity, "audiorecord.pcm")
+
+    suspend fun onStopped() {
+        appStateMutatorMutex.mutate {
+            playbackState = PlaybackState.Ready
+        }
+    }
+
+    suspend fun onMicClicked() {
+        appStateMutatorMutex.mutate {
```
manuelvicnt, line 91: Holding the mutex at this level seems unsafe. Given that in this block you call suspend functions (e.g. `record` in the first block), it might be the case that you hold the mutex for too long. Is that the intention? Is the intention that while `onMicClicked` is being executed, other functions cannot update the state?

Maybe the functions need to be called differently? `onMicClicked` is not self-explanatory, it just tell us that it's handling a user event. Then I have to go through the body to actually see what's going on.

manuelvicnt: On the other hand, it makes sense that the functions are named based on the input they're processing as this is a state holder that's tightly coupled to the UI. Up to you!

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainState.kt
```diff
@@ -0,0 +1,274 @@
...
+    private val activity: Activity,
+    private val requestPermission: () -> Unit,
+    private val showPermissionRationale: () -> Unit,
+    private val showSpeakerNotSupported: () -> Unit
+) {
+    /**
+     * The [MutatorMutex] that guards the state of the app.
+     *
+     * Due to being a [MutatorMutex], this automatically handles cleanup of any ongoing asynchronous work,
+     * like playing music or recording, ensuring that only one operation is occurring at a time.
```
manuelvicnt, line 54: Can you also explain here what happens when an interruption happens? IIRC if the priorities are the same, the previous work is interrupted. Is that right?

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainStateHolder.kt
manuelvicnt: SGTM, thanks for explaining! Let's add more documentation around that as comments :D it's a very important behavior to keep in mind

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
manuelvicnt: The new implementation seems easier to understand thank you! I added an additional comment to that impl that I think could be even simpler. Thank you

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerScreen.kt
alexvanyo: Updated!

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
alexvanyo: Replaced with a `DisposableEffect` for setting up a `LifecycleObserver`

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainState.kt
alexvanyo: It's a `private val`, and a `MutatorMutex` is a different type from a normal `Mutex`, so I think the name makes sense to indicate the special type. I did rename to `playbackStateMutatorMutex` to be more in line with the new `PlaybackState` name.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainState.kt
alexvanyo: Done!

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerApp.kt
manuelvicnt: Thanks! Looks good

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/SpeakerScreen.kt
manuelvicnt: Sorry, I should've explained myself better. I meant remembering the `derivedStateOf` function. Something like

```
= remember {
    derivedStateOf { ... }
}
```

I believe that's more efficient than doing `remember(x,y)` but I'm not 100% sure, maybe we need to check with eng to see what's preferable?

manuelvicnt: Also, I've seen that we have a usage of `derivedStateOf` in Jetnews which isn't remembered. We might need to fix that in the same way. Thanks!

alexvanyo: Per chats about `derivedStateOf` and `remember`, I'm removing it entirely from these usages.

The computations are simply enough that there's no need for overcomplicating them, and using `derivedStateOf` properly is surprising subtle.

I kept the extracted function, but it's now a plain old Kotlin function that computes the ui state.

alexvanyo:
Hi @manuelvicnt ! I got this updated with the latest APIs, and removed `derivedStateOf` entirely to keep everything simple.

@codingjeremy using the latest Compose Wear release, I am using the new `AlertDialog` composables here now instead.

yschimke (approved):
As an inplace conversion it seems good.  There is arguably a lot of complexity, but that in some way helps the sample because it shows a lot of advanced concepts, for compose and coroutines.

README.md
```diff
@@ -31,7 +31,7 @@ Samples
 
 * **[WearOAuth](WearOAuth)** (Kotlin) - Demonstrates how developers can authenticate a user on their Wear OS app via the user's mobile/phone device without requiring a mobile app (Wear OS companion app handles the request on the mobile side). The sample uses OAuth. [Guide](https://developer.android.com/training/wearables/apps/auth-wear)
 
-* **[WearSpeakerSample](WearSpeakerSample)** (Kotlin) - Demonstrates audio recording and playback if the wearable device has a speaker. [Guide](https://developer.android.com/training/wearables/wearable-sounds)
+* **[WearSpeakerSample](WearSpeakerSample)** (Compose/Kotlin) - Demonstrates audio recording and playback if the wearable device has a speaker. [Guide](https://developer.android.com/training/wearables/wearable-sounds)
```
codingjeremy, line 34: I would call out this is an advanced sample that uses effects and ConstraintLayout and link both of those to the guides. That way they know before even getting to the main README what it covers.

WearSpeakerSample/README.md
```diff
@@ -1,11 +1,16 @@
...
+The Compose UI is fairly advanced, using [ConstraintLayout for Compose](https://developer.android.com/jetpack/compose/layouts/constraintlayout).
```
codingjeremy, line 11: Actually, can you also call out this uses several effects APIs and link to the guide? That's pretty important too.

codingjeremy (commented):
Overall, it looks good. I'm not an expert on the effects APIs, so I'm glad Manuel and Jolanda covered that.

The one main problem I see is there are lots of Column limit style violations. The [Kotlin Style Guide calls for 100 char limit](https://engdoc.corp.google.com/eng/doc/devguide/java/style/index.md?cl=head#s4.4-column-limit) and that's the same as the internal Java style guide.

I was going to mark all the violations, but I wasn't sure if you did this for a reason or I missed something?

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainState.kt
```diff
@@ -0,0 +1,286 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
+import java.time.Duration
+import kotlin.coroutines.resume
+
+/**
+ * A state holder driving the logic of the app.
+ *
+ * This state holder is scoped to the composition (and therefore can be provided an [Activity]) to handle
```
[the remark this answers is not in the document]
alexvanyo: Updated!

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainState.kt
```diff
@@ -0,0 +1,286 @@
...
+}
+
+/**
+ * Plays the MP3 file embedded in the application, updating the state to reflect the playing.
+ */
+private suspend fun playMusic(activity: Activity) {
+    val mediaPlayer = MediaPlayer.create(activity, R.raw.sound)
+
+    try {
+        suspendCancellableCoroutine<Unit> { cont ->
```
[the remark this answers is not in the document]
alexvanyo: Added!

README.md
alexvanyo: Done!

WearSpeakerSample/README.md
alexvanyo: Done!

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/MainState.kt
```diff
@@ -0,0 +1,286 @@
...
+            playbackState = PlaybackState.Ready
+        }
+    }
+
+    suspend fun onMicClicked() {
+        playbackStateMutatorMutex.mutate {
+            when (playbackState) {
+                is PlaybackState.Ready,
+                PlaybackState.PlayingVoice,
+                PlaybackState.PlayingMusic ->
```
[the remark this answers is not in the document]
alexvanyo: Yes, although it's pretty difficult to trigger. If you tap multiple buttons at the same time, each one's click listener will still get called, so if it goes `onMusicClicked` and then `onMicClicked`, the state could be `PlayingMusic` when the `onMicClicked` interrupts playing music to try to record.

codingjeremy (approved):

alexvanyo:
Thanks everyone for the reviews and feedback!

PR #161, opened 2021-10-11
Adds a Compose sample using Scaffold, Navigation, Curved Text, and more...

codingjeremy:
Tried to keep as simple as possible, PTAL.


yschimke:
> Adding in followup PR.

Thanks, and if it balloons out let me know, happy to help. I think it should be straightforward and help a lot with an encouraging sample for adding good instrumentation tests on compose UIs.

PR #163, opened 2021-10-13
Updates dependencies.

codingjeremy:
Updates to latest dependencies. 

@LouisCAD - FYI


codingjeremy:
> Would it make sense to use TODO comments instead of regular comments for the ui-tooling version?

Yeah, that makes sense; added.

PR #173, opened 2021-10-26
[AlwaysOn] Use Gradle Managed Devices for testing

alexvanyo:
Note: currently blocked on https://issuetracker.google.com/issues/204590177

### DIFF
--- a/.github/workflows/AlwaysOnKotlin.yml
+++ b/.github/workflows/AlwaysOnKotlin.yml
@@ -61,8 +61,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        profile: [ wear_round, wear_square ]
-        api-level: [ 26, 28, 30 ]
+        device:
+          - roundApi26
+          - roundApi28
+          - roundApi30
+          - squareApi26
+          - squareApi28
+          - squareApi30
 
     steps:
       - name: Checkout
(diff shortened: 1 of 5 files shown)


.github/workflows/AlwaysOnKotlin.yml
```diff
@@ -58,11 +58,6 @@ jobs:
     needs: build
     runs-on: macOS-latest # enables hardware acceleration in the virtual machine
     timeout-minutes: 30
-    strategy:
-      fail-fast: false
-      matrix:
-        profile: [ wear_round, wear_square ]
-        api-level: [ 26, 28, 30 ]
```
yschimke, line 65: Losing the matrix here is unfortunate as the entire job will pass or fail? Can you retain the matrix with the device names?  Will be easy to check which devices are failing, and possibly faster?

alexvanyo: Good idea! (having them done in one job will use only a single VM instead of 6, but that's probably fine)
 
 All of them are blocked on https://issuetracker.google.com/issues/204590177 right now.

yschimke:
Will close and reopen once the bug upstream lands. Merge conflicts and still blocked.

yschimke:
@alexvanyo This is unblocked, but only if we adopt AGP 7.3 alphas.  Will hold off for now.

alexvanyo:
@yschimke maybe worth reopening to check that everything is passing GitHub Actions?

And not merge until stable. That could verify everything works for CI, without bumping AGP for everyone yet.

PR #180, opened 2021-11-01
[WatchFace] Update to 1.0.0-beta01

alexvanyo:
Updates WatchFaceAlphaKotlin to 1.0.0-beta01 (the latest release). The API shifted slightly, so I adjusted for that and also simplified a few things that I noticed.

Also, `WatchFaceAlphaKotlin` probably isn't the best name, since it's now in beta 😄


codingjeremy (approved):
Thanks for cleaning up!

I just had some minor notes on comments I don't think I wrote very clearly. 

Feel free to edit them, but it might help devs from having to think through a lot of the calculations.

codingjeremy (approved):
Perfect, thanks!

codingjeremy:
Added @ADClarke as FYI

codingjeremy:
Talked with Alex C and we are good to go, thanks!

PR #186, opened 2021-11-04
Delete old WatchFace sample

alexvanyo:
This removes the old WatchFace sample in preparation to renaming the WatchFaceAlphaKotlin sample. I also updated the README to remove references to samples that will be gone.


codingjeremy (approved):
Before submitting, can you sync with Anna to make sure they are removed from DAC?

PR #194, opened 2021-12-01
Update to kotlinx.coroutines 1.6.0

alexvanyo:
Updates AlwaysOn to kotlinx.coroutines 1.6.0, and also updates the tests to use the new kotlinx.coroutines test API.

### DIFF
--- a/AlwaysOnKotlin/Wearable/src/androidTest/java/com/example/android/wearable/wear/alwayson/MainActivityTests.kt
+++ b/AlwaysOnKotlin/Wearable/src/androidTest/java/com/example/android/wearable/wear/alwayson/MainActivityTests.kt
@@ -32,8 +32,10 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.platform.app.InstrumentationRegistry
 import androidx.test.uiautomator.UiDevice
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import kotlinx.coroutines.runBlocking
-import kotlinx.coroutines.test.TestCoroutineDispatcher
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.TestCoroutineScheduler
+import kotlinx.coroutines.test.TestScope
+import kotlinx.coroutines.test.runTest
 import org.junit.After
 import org.junit.Before
 import org.junit.Test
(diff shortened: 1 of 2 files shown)


AlwaysOnKotlin/Wearable/src/androidTest/java/com/example/android/wearable/wear/alwayson/MainActivityTests.kt
```diff
@@ -99,10 +100,12 @@ class MainActivityTests {
             )
         onView(withId(R.id.state)).check(matches(withText(context.getString(R.string.mode_active_label))))
         onView(withId(R.id.draw_count)).check(matches(withText(context.getString(R.string.draw_count_label, 1))))
+
+        scenario.moveToState(Lifecycle.State.STARTED)
```
alexvanyo, line 104: If we don't pause the activity, then `runTest` fails with uncompleted work. This makes sense: we are always enqueuing an infinite amount of work, so we need to cancel it before finishing `runTest` by pausing the activity.

alexvanyo:
Hi @JoseAlcerreca and @manuelvicnt , I updated a simple sample here to use the new kotlinx-coroutines-test APIs. This was a pretty straightforward change here, since I wanted fine control over timing due to the sample under test.

AlwaysOnKotlin/Wearable/src/androidTest/java/com/example/android/wearable/wear/alwayson/MainActivityTests.kt
codingjeremy: I'm confused, should there be some call to pause it? I'm new to this API though, so I might be missing something. (If so, ignore this.) :)

Also, should we use your comment as a comment in the code?

codingjeremy (approved):
Just one comment to your comment.

AlwaysOnKotlin/Wearable/src/androidTest/java/com/example/android/wearable/wear/alwayson/MainActivityTests.kt
alexvanyo: `moveToState` `STARTED` will move from `RESUMED` to `STARTED`, thereby pausing it. I'll add my comment inline to make that clear!

alexvanyo: Updated!

codingjeremy: Perfect, thanks!

PR #199, opened 2021-12-13
[DataLayer] Mobile update

alexvanyo:
Updates the mobile side of the DataLayer sample to Kotlin and Compose.

This is a pretty simple update, as the data layer implementation could be simplified with a `ViewModel` and coroutines.


codingjeremy (approved):
LGTM, thanks for doing this! Ran really well on my physical devices.

PR #200, opened 2021-12-15
[DataLayer] Wear update

alexvanyo:
Updates the Wear side of the `DataLayer` sample to use Kotlin, coroutines and Compose.


DataLayer/Application/src/main/java/com/example/android/wearable/datalayer/MainActivity.kt
```diff
@@ -128,13 +129,39 @@ class MainActivity : ComponentActivity() {
             Uri.parse("wear://"),
             CapabilityClient.FILTER_REACHABLE
         )
+
+        if (isCameraSupported) {
+            lifecycleScope.launch {
+                try {
+                    capabilityClient.addLocalCapability(CAMERA_CAPABILITY).await()
```
alexvanyo, line 136: @codingjeremy I added the dynamic camera capability here per our earlier discussion, it wasn't too hard to add!

codingjeremy (approved):
I'm so confused on this PR... I thought I made comments on this but they don't show up in the history?

Did you do a force push vs. a new commit?

I'm pretty sure you resolved everything from a quick review but can't remember. Is that correct?

alexvanyo:
I don't think I force pushed since I changed it from a draft to a full PR. This is on the wear side, whereas #199 was for the mobile side.

PR #204, opened 2021-12-30
Adds support for multiple scrolling screens and proper support for Vignette.

codingjeremy:
The details screen now support scrolling, so the sample shows two different types of scrolling for one Scaffold. 

Let me know if this is the right design you had in mind.

I also added the proper vignette support.


ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -162,10 +198,27 @@ fun WearApp(watchRepository: WatchRepository) {
...
         }
     }
 }
+
+/**
+ * Used for the 'in' comparisons in the Scaffold section for determining which screen is visible
+ * and what to show (vignette and position indicator). Note, 'in' is translated to the operator
+ * 'contains' fun in Kotlin.
+ */
+operator fun String?.contains(substring: String): Boolean {
```
alexvanyo, line 215: This seems a little strange to me, why not just use equals?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -73,44 +76,77 @@ class MainActivity : ComponentActivity() {
...
-                )
+                // Only show vignette for screens with scrollable content.
+                if (Screen.WatchList.route in currentRoute) {
+                    CustomVignette(
+                        visible = vignetteVisiblePreference,
+                        vignettePosition = VignettePosition.TopAndBottom
+                    )
+                } else if (Screen.WatchDetail.route in currentRoute) {
+                    CustomVignette(
+                        visible = true,
```
alexvanyo, line 131:
```suggestion
                        visible = vignetteVisiblePreference,
```

Did you want this to also use `vignetteVisiblePreference`?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -73,44 +76,77 @@ class MainActivity : ComponentActivity() {
...
+        // scroll indicator).
+        // In our case, the watch list screen uses a ScalingLazyColumn and the watch details
+        // screen uses a Column with scrolling enabled.
+        // We also use the states for various other things (like hiding the time when the user is
+        // scrolling).
+        // ----------
+        // State for the ScalingLazyColumn (watch list screen)
         val scalingLazyListState: ScalingLazyListState = rememberScalingLazyListState()
+        // State for the Column's vertical scroll modifier (watch detail screen).
+        val watchDetailScrollState: ScrollState = rememberScrollState()
```
alexvanyo, line 90: One interesting effect of this setup: Because this `ScrollState` is hoisted outside of the navigation, this state will be preserved even when the user leaves the detail to go back. That causes the following:

- User navigates to the list
- User navigates to detail A
- User scrolls to the bottom of detail A
- User navigates back
- User navigates to detail B
- User is now already scrolled on detail B with detail A's scroll state.

Avoiding that is a bit tricky, and in the general case with nested navigation (say, if you could navigate from a detail to another detail view) since we can't just have a nice, fixed set of hoisted states.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: Not in this case, I always want the vignette to appear here. 

The only reason I let the user toggle it off/on in the list page was so they could see the visual difference... Usually we don't want them turning it off/on at all in an app... I wonder if I should just remove that feature to avoid confusion.

WDYT?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: Yep, that's why.

I used the `contains` based on the code in Jetsnack, but, yeah, I think no extension fun with just `startsWith` is probably clearer (especially for folks who aren't used to the `in` + `operator contains` convention).

Changed, PTAL.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: Yeah, not ideal... I guess I could reset the scroll state value to 0 for each detail screen launch or do some sort of logic where if the detail screen isn't equal to the previous detail screen that was used it resets.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
alexvanyo: Looks good!

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
alexvanyo: I could go either way, but yeah I think removing anyway to hide it on a scrolling screen makes sense.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
alexvanyo: Yeah, either of those should work, although it's unfortunate it doesn't solve the general case... It's unfortunate that `NavController.backQueue` is hidden.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: My understanding though in general is we shouldn't remove the vignette on scroll. @JohnNichol is that correct?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: I reworked it to use a remembered map and that works pretty well. PTAL.

codingjeremy:
@JohnNichol @BowerSteve - Mind taking a look at [the latest commit](https://github.com/android/wear-os-samples/pull/204/commits/7f96a1d3cd71986ae0d38a7c515060909ac1d0a0) in this PR?

I dynamically handle all the scroll states by using a remembered map. It works pretty well, but I wanted to verify that's what you had in mind. Thanks!

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -71,44 +74,96 @@ class MainActivity : ComponentActivity() {
...
-                CustomPositionIndicator(
-                    visible = scalingLazyListState.isScrollInProgress,
-                    scalingLazyListState = scalingLazyListState
-                )
+                // Only show position indicator for scrollable content.
+                currentRoute?.let {
+                    val activelyScrolling =
+                        scrollStates[currentScrollStateKey]?.isScrollInProgress ?: false
+
+                    if (currentRoute.startsWith(Screen.WatchList.route)) {
```
yschimke, line 151: This looks like it's simpler to just use the runtime type information to make the decision.  Or put this logic inside of CustomPositionIndicator so it takes a ScrollableState?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -71,44 +74,96 @@ class MainActivity : ComponentActivity() {
...
+        // ----------
+
+        // For all scrolling screens, we save their scrolling offset in a remembered map (as a
+        // ScrollableState), so when users return to the screen, they are at the same location as
+        // they left.
+        // We set the default to the watch list screen (ScalingLazyListState) since there is only
+        // one. For the watch detail screens (with many different variations based on the watch id),
+        // we add those dynamically to the map as they are needed (in the navigation section).
+        var currentScrollStateKey: String = remember { Screen.WatchList.route }
+        val scrollStates: MutableMap<String, ScrollableState> = remember {
```
yschimke, line 93: My concern about this is it looks like the kind of logic we expect ultimately every app with navigation to implement. I'd argue we should push to move this logic, and any logic for showing/hiding or scaling, offsettting the TimeText into Scaffold.

cc @JohnNichol

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -71,44 +74,96 @@ class MainActivity : ComponentActivity() {
...
-        // Allows user to disable the text before the time and hide the vignette.
+        // Allows user to disable the text before the time.
         var showProceedingTextBeforeTime by rememberSaveable { mutableStateOf(false) }
-        var vignetteVisible by rememberSaveable { mutableStateOf(true) }
+
+        // Allows user to show/hide the vignette on appropriate screens.
+        // IMPORTANT NOTE: Usually you want to show the vignette all the time on screens with
+        // scrolling content, a rolling side button, or a rotating bezel. This preference is just
+        // to visually demonstrate the vignette for the developer to see it on and off.
+        var vignetteVisiblePreference by rememberSaveable { mutableStateOf(true) }
```
yschimke, line 106: Can the Vignette be driven off the ScrollableState? e.g. enabled for ScalingLazyListState when we know there is enough content to cause scrolling?

yschimke:
It's a great demo for doing advanced things in Wear Compose.  But I fear it shows patterns that the UX team wouldn't necessarily advise following, and where they do we should arguably make more of this the default in a Navigation aware Scaffolding...

yschimke (approved):
It's a good advanced sample, and I'd be for incrementally addressing further comments.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: +1

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: Hmmm... I suppose we could in a followup PR. You mean, if it is some sort of scrolling window BUT it doesn't have enough content to scroll, don't show the Vignette?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: Ah, good call, PositionIndictor takes care of this... maybe it didn't when I first wrote the sample? I can't remember. Either way, let me get rid of this and just use the stock PositionIndicator since it handles hiding itself.

codingjeremy: Fixed, PTAL

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: Created a bug: https://issuetracker.google.com/issues/217446791

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -71,44 +74,91 @@ class MainActivity : ComponentActivity() {
...
+        // scrolling).
+        // ----------
+
+        // For all scrolling screens, we save their scrolling offset in a remembered map (as a
+        // ScrollableState), so when users return to the screen, they are at the same location as
+        // they left.
+        // We set the default to the watch list screen (ScalingLazyListState) since there is only
+        // one. For the watch detail screens (with many different variations based on the watch id),
+        // we add those dynamically to the map as they are needed (in the navigation section).
+        var currentScrollStateKey: String = remember { Screen.WatchList.route }
```
alexvanyo, line 92: This line seems very suspicious to me.

There isn't a `mutableStateOf` here, so any time `currentScrollStateKey` is set, it will just be wiped out with next recomposition.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -156,10 +210,18 @@ fun WearApp(watchRepository: WatchRepository) {
                         }
                     )
                 ) { navBackStackEntry ->
-                    val watchId =
-                        navBackStackEntry.arguments?.getInt(WATCH_ID_NAV_ARGUMENT)!!
+                    val watchId = navBackStackEntry.arguments?.getInt(WATCH_ID_NAV_ARGUMENT)!!
+                    currentScrollStateKey = Screen.WatchDetail.route + watchId
```
alexvanyo, line 214: This is a side-effect of composition, which is something in general to avoid. I think the reason why this currently works is that this value is always set right before the `PositionIndicator` is reading it, so it queries for the correct value.

alexvanyo (changes requested):

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: Oh, good catch, still getting used to the side effect stuff. Let me rewrite it with those APIs.

codingjeremy: I tried a bunch of different variations that worked, but I tired out the back stack approach you outlined and it worked without the FR you put it. :)

As discussed, we can get sign off from Ian, but I think that's the best solution.

codingjeremy:
TL;DR: After a bunch of different solutions, I settled on building out a map of the scrollable states based on the back stack (Alex's suggestion). This actually works even without the FR Alex put in. 

We just need to verify with Ian it is ok to move forward with that and I can commit (after you all approve). Thanks for your patience!

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
yschimke: I actually meant collapsing the current if {} else {} and getting the scrollState[currentId].  Then having an if else based on the type.

My thinking was this logic needs to change whenever routes change, while checking based on types you can even implement all 4 forms of PositionIndicator now and never change when routes changes.

alexvanyo (changes requested):
Ian responded in https://issuetracker.google.com/issues/217465473, with a better approach that doesn't use the `backQueue`, and even though that "works" it isn't a supported usage and it wouldn't work in general with multiple backstacks.

yschimke:
> Ian responded in https://issuetracker.google.com/issues/217465473, with a better approach that doesn't use the `backQueue`, and even though that "works" it isn't a supported usage and it wouldn't work in general with multiple backstacks.

I like this suggestion using arguments to save the scroll type.

codingjeremy:
It's now using the solution Ian recommended.

yschimke (approved):
Love how this ended up

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -132,36 +201,64 @@ fun WearApp(watchRepository: WatchRepository) {
...
+                        // as information for the destination.
+                        navArgument(SCROLL_TYPE_NAV_ARGUMENT) {
+                            type = NavType.EnumType(DestinationScrollType::class.java)
+                            defaultValue = DestinationScrollType.SCALING_LAZY_COLUMN_SCROLLING
                         }
                     )
+                ) {
+                    val passedScrollType = it.arguments?.getSerializable(SCROLL_TYPE_NAV_ARGUMENT)
+
+                    if (passedScrollType == DestinationScrollType.SCALING_LAZY_COLUMN_SCROLLING) {
```
alexvanyo, line 217: Nit: Do we need the `if` statements here inside the `composable`? It might make sense to do a `check(passedScrollType == DestinationScrollType.SCALING_LAZY_COLUMN_SCROLLING)` if there's no reason it should ever be a different one versus a strange blank screen.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -71,44 +75,109 @@ class MainActivity : ComponentActivity() {
...
+        // We also use these scrolling states for various other things (like hiding the time
+        // when the user is scrolling and only showing the vignette when the screen is
+        // scrollable).
+        //
+        // Remember, mobile guidelines specify that if you back navigate out of a screen and then
+        // later navigate into it again, it should be in its initial scroll state (not the last
+        // scroll location it was in before you backed out).
+
+        // TODO: Replace with .currentBackStackEntryAsState() once added to Compose for Wear OS
+        // Navigation Library: https://issuetracker.google.com/issues/212739653
```
alexvanyo, line 112: Nit: I think `alpha15` added this?

alexvanyo (approved):
Looks good, just a couple of nits!

I also like how this ended up here.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: Ah, yes, good catch, fixed.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: Yep, you are right, I missed it. Fixed.

alexvanyo (approved):

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
JohnNichol: Correct - My understanding of the UX spec is that Vignette is always expected to be in place for scrolling content.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
JohnNichol: @codingjeremy the use of the map looks fine to me. 

I don't think this logic should be moved into the Scaffold - the Scaffold doesn't know about navigation or app structure. If we start to move things into the Scaffold then it needs to start to expose all-sorts of the detail back via it's state to allow the application to manage things like transitions. We spent a lot of time with API Sponsors early on workout out whether the Scaffold knew about Vignettes, PositionIndicators, etc. and came to the conclusion that it made the Scaffold inflexible and difficult to use. 

It's a reasonable question to consider whether there is a component/components that contain a Vignette and provide this functionality for reuse in different application contexts.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
```diff
@@ -71,44 +75,105 @@ class MainActivity : ComponentActivity() {
...
+                when (scrollType) {
+                    DestinationScrollType.SCALING_LAZY_COLUMN_SCROLLING -> {
+                        CustomVignette(
+                            visible = vignetteVisiblePreference,
+                            vignettePosition = VignettePosition.TopAndBottom
+                        )
+                    }
+                    DestinationScrollType.COLUMN_SCROLLING -> {
+                        CustomVignette(
+                            visible = true,
```
BowerSteve, line 154: Should this be vignetteVisiblePreference?

JohnNichol (approved):
Overall looking good - suggest we merge and iterate.

BowerSteve (approved):
LGTM - left one minor comment about the vignetteVisibility.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
yschimke: Ignore, think we can do this outside the sample.  I'll discuss with you.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/MainActivity.kt
codingjeremy: I was initially keeping them separate so it just showed on the list page, but I can see what you mean, i.e., it may be more confusing why they are separated. I combined them to simplify them, thanks!

PR #206, opened 2022-01-19
Update to Wear Specific Previews

yschimke:
Attempting to upgrade to Wear specific Previews where appropriate (full screen).

### DIFF
--- a/ComposeAdvanced/app/build.gradle
+++ b/ComposeAdvanced/app/build.gradle
@@ -42,6 +42,8 @@ android {
     kotlinOptions {
         jvmTarget = '1.8'
         useIR = true
+        freeCompilerArgs += "-opt-in=androidx.wear.compose.material.ExperimentalWearMaterialApi"
+        freeCompilerArgs += "-opt-in=androidx.compose.animation.ExperimentalAnimationApi"
     }
 
     buildFeatures {
--- a/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/components/CustomPositionIndicator.kt
+++ b/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/components/CustomPositionIndicator.kt
@@ -50,6 +50,7 @@ fun CustomPositionIndicator(
     showSystemUi = true,
     device = Devices.WEAR_OS_SQUARE
 )
+// This will only be rendered properly in AS Chipmunk and beyond
 @Preview(
     apiLevel = 26,
     uiMode = Configuration.UI_MODE_TYPE_WATCH,
--- a/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/components/CustomPositionIndicator.kt
+++ b/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/components/CustomPositionIndicator.kt
@@ -17,6 +17,7 @@ package com.example.android.wearable.composeadvanced.presentation.components
 
 import android.content.res.Configuration
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.tooling.preview.Devices
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.wear.compose.material.PositionIndicator
 import androidx.wear.compose.material.ScalingLazyListState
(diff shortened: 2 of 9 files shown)


ComposeAdvanced/app/build.gradle
```diff
@@ -42,6 +42,8 @@ android {
     kotlinOptions {
         jvmTarget = '1.8'
         useIR = true
+        freeCompilerArgs += "-opt-in=androidx.wear.compose.material.ExperimentalWearMaterialApi"
```
yschimke, line 45: Can move or split this out, but for a project adopting Wear Compose, putting these in every Composable seems overkill.  Should opt-in once.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/components/CustomPositionIndicator.kt
```diff
@@ -38,12 +39,22 @@ fun CustomPositionIndicator(
...
+    apiLevel = 26,
+    uiMode = Configuration.UI_MODE_TYPE_WATCH,
+    showSystemUi = true,
+    device = Devices.WEAR_OS_SQUARE
+)
+@Preview(
+    apiLevel = 26,
+    uiMode = Configuration.UI_MODE_TYPE_WATCH,
+    showSystemUi = true,
+    device = Devices.WEAR_OS_SMALL_ROUND
```
yschimke, line 57: Note: this requires the Android Studio Canary.  Is it safe to assume that given we are on Alpha Compose versions?

We could also (probably) warn in the the gradle build if you are using an older Android Studio.

alexvanyo: A crude-ish way to force that would be to update the AGP to the `7.2.x` alpha corresponding to Chipmunk.

That's not super friendly though, as most people would probably be on stable.

yschimke: It shouldn't break the build, but the preview won't be clean, but getting developers familiar with this syntax sounds nice.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/components/WatchAppChip.kt
```diff
@@ -83,19 +82,18 @@ fun WatchAppChip(
...
 )
 @Composable
 fun PreviewWatchAppChip() {
-    WatchAppChip(
-        watchModelNumber = 123456,
-        watchName = "Watch 123456 Name",
-        watchIcon = R.drawable.ic_watch,
-        onClickWatch = { }
-    )
+    Column {
```
alexvanyo, line 90: Just curious, what does the Column here do?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/components/CustomPositionIndicator.kt
alexvanyo: I think a comment next to the `device` field with "This will only be rendered properly in AS Chipmunk and beyond" would be good to call that out while still making it easy to sync and build the project with the current stable AS.

Once Chipmunk becomes stable we can remove all of those then.

alexvanyo (approved):
Looks good to me!

codingjeremy:
Hi all, 

This is a cool feature but should we be pushing the main version of compose to an alpha from a RC (almost stable)?

I noticed it was for a beta studio feature anyway too? Thoughts?

I know you could argue Compose for Wear OS is in alpha so an alpha for compose itself as well isn't a big deal but I thought it might be better for devs to see we rely on a production version (or close to it) of Compose itself. Might help with the larger partners who want more stability.

Maybe not, but I thought I would asked as I noticed this change as I was finishing up my [PR](https://github.com/android/wear-os-samples/pull/204). Thanks!

yschimke:
@codingjeremy But Wear Compose is moving to the Alpha also.  It's the required version.  This PR doesn't change that.

See https://mvnrepository.com/artifact/androidx.wear.compose/compose-foundation/1.0.0-alpha15

PR #212, opened 2022-02-16
Update AGP and Compose versions

shumelchyk:


alexvanyo:
Just a note, it would probably make sense to update AGP and the Compose version across all samples and not just `ComposeAdvanced` (there isn't a central place to do that, just in each project individually)

shumelchyk:
Good point, I can add dependencies updates for other samples within current PR. Moving forward - do we have any suggested best practices how to bump dependencies nicely across all repos (otherwise I'll investigate)?

yschimke:
@shumelchyk this is the way https://github.com/android/wear-os-samples/issues/213

shumelchyk:
https://github.com/android/wear-os-samples/issues/213 looks promising, let's try to integrate it, thanks!

PR #215, opened 2022-02-18
Update advanced dependencies

yschimke:
Removes the automated plugins, so we can adopt something like the renovate process.


shumelchyk (approved):
LGTM, thanks for moving quickly :)

PR #221, opened 2022-02-22
Always on emulator fix

yschimke:


shumelchyk:
Checks seems to fail anyway with the updated workflow?

yschimke:
@shumelchyk I'd like to land this as I think it moves us closer.

shumelchyk (approved):
LGTM

PR #224, opened 2022-02-24
Build all modules when dependency versions change.

yschimke:
Should land before https://github.com/android/wear-os-samples/pull/223


shumelchyk (approved):
LGTM

PR #226, opened 2022-03-01
Cleanup and apply spotless to all files

yschimke:
This would not be a priority for me, but we seem to spend a lot of time in PRs on line length, so figured it would be good to cleanup once and for all.


luizgrp (approved):
LGTM - thanks for doing this 👍

PR #229, opened 2022-03-03
Use Main dispatcher in Tiles samples

yschimke:
Run tileRequest and resourceRequest on Dispatchers.Main.

This avoid an observable behaviour change from TileService.

### DIFF
--- a/WearTilesKotlin/app/src/main/java/com/example/wear/tiles/CoroutinesTileService.kt
+++ b/WearTilesKotlin/app/src/main/java/com/example/wear/tiles/CoroutinesTileService.kt
@@ -34,7 +34,7 @@ abstract class CoroutinesTileService : TileService() {
     private val serviceJob = Job()
     protected val serviceScope = CoroutineScope(Dispatchers.Main + serviceJob)
 
-    override fun onTileRequest(
+    final override fun onTileRequest(
         requestParams: TileRequest
     ): ListenableFuture<Tile> = serviceScope.future {
         tileRequest(requestParams)
@@ -42,15 +42,16 @@ abstract class CoroutinesTileService : TileService() {
 
     abstract suspend fun tileRequest(requestParams: TileRequest): Tile
 
-    override fun onResourcesRequest(requestParams: ResourcesRequest):
+    final override fun onResourcesRequest(requestParams: ResourcesRequest):
         ListenableFuture<Resources> = serviceScope.future {
             resourcesRequest(requestParams)
         }
 
-    open suspend fun resourcesRequest(requestParams: ResourcesRequest): Resources =
-        Resources.Builder().setVersion(FIXED_RESOURCES_VERSION).build()
+    abstract suspend fun resourcesRequest(requestParams: ResourcesRequest): Resources
 
-    companion object {
-        private const val FIXED_RESOURCES_VERSION = "1"
+    override fun onDestroy() {
+        super.onDestroy()
+        // Cleans up the coroutine
+        serviceJob.cancel()
     }
 }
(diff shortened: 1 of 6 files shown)


WearTilesKotlin/app/src/main/java/com/example/wear/tiles/CoroutinesTileService.kt
```diff
@@ -0,0 +1,56 @@
...
+        tileRequest(requestParams)
+    }
+
+    abstract suspend fun tileRequest(requestParams: TileRequest): Tile
+
+    override fun onResourcesRequest(requestParams: ResourcesRequest):
+        ListenableFuture<Resources> = serviceScope.future {
+            resourcesRequest(requestParams)
+        }
+
```
alexvanyo, line 49: Is the `onDestroy` callback to cancel the scope missing here?

gradle/libs.versions.toml
```diff
@@ -26,6 +26,7 @@ androidx-datastore-preferences = "androidx.datastore:datastore-preferences:1.0.0
 androidx-fragment-ktx = "androidx.fragment:fragment-ktx:1.4.1"
 androidx-lifecycle-common-java8 = { module = "androidx.lifecycle:lifecycle-common-java8", version.ref = "androidx-lifecycle" }
 androidx-lifecycle-runtime-ktx = { module = "androidx.lifecycle:lifecycle-runtime-ktx", version.ref = "androidx-lifecycle" }
+androidx-lifecycle-service = { module = "androidx.lifecycle:lifecycle-service", version.ref = "androidx-lifecycle" }
```
alexvanyo, line 29: Was this used anywhere? I don't see it being imported in `build.gradle`.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/CoroutinesTileService.kt
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Base class for a Kotlin and Coroutines friendly TileService.
+ */
+abstract class CoroutinesTileService : TileService() {
+    // For coroutines, use a custom scope we can cancel when the service is destroyed
+    private val serviceJob = Job()
+    protected val serviceScope = CoroutineScope(Dispatchers.Main + serviceJob)
+
+    override fun onTileRequest(
```
alexvanyo, line 37: It might be worth making this and `onResourcesRequest` `final override fun`, as subclasses should just implement the other methods.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/CoroutinesTileService.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    abstract suspend fun tileRequest(requestParams: TileRequest): Tile
+
+    override fun onResourcesRequest(requestParams: ResourcesRequest):
+        ListenableFuture<Resources> = serviceScope.future {
+            resourcesRequest(requestParams)
+        }
+
+    open suspend fun resourcesRequest(requestParams: ResourcesRequest): Resources =
+        Resources.Builder().setVersion(FIXED_RESOURCES_VERSION).build()
```
alexvanyo, line 51: It doesn't looks like this default implementation is used anywhere, maybe it'd be best to just have this be an `abstract` method as well?

alexvanyo (changes requested):
Overall looks good, just a few small things!

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/CoroutinesTileService.kt
yschimke: Ooops, good catch.

PR #231, opened 2022-03-04
Quick fix for samples

### LINKED ISSUE #230: Wear OS Samples from Android Studio are broken
I am working in Android Studio, and I followed the directions here: https://developer.android.com/training/wearables/compose-setup 

When it was done downloading, I got an error that states: "In version catalog libs, import of external catalog file failed" basically it looks like it might be looking in the wrong directory for the file. The support link that it mentions just says to make sure I have the file.

Anyone have a way I can set the directory to the right file or do I need to download something else before trying the sample app?
Also if this is not the right place to ask for help pls direct me to where I may find help

yschimke:
Fix for importing samples which broke here https://github.com/android/wear-os-samples/pull/220

fixes #230


yschimke:
I'll follow up with a cleanup/minimisation of the libs.versions.toml files.  But want to fix the samples first.

yschimke:
I've logged https://github.com/android/wear-os-samples/issues/232 to avoid this happening again, sorry!

yschimke:
@codingjeremy @alexvanyo Going to land to test the fix in Android Studio.  Will follow up with a cleanup of dependencies, since each is the union of all.

Issue #238, opened 2022-03-12
Jank and poor scrolling performance on the ComposeAdvanced sample


LouisCAD:
Hello,

I just tried the current latest ComposeAdvanced sample on my both watches, a TicWatch E3 and a Samsung Galaxy Watch 4, and scrolling the entire app is stuttering all-the-time:
- First, a cold start is quite slow. 5 seconds for first non black pixel, 6 seconds for a proper render. That's after manipulating the app beforehand then "killing" it since Samsung's "skin" allows it, so the ART profile should be present.
- Scrolling the list under the "List of Watches" renders at like 10 frames per second, sometimes less.
- Flings are not always honored, quite often, the gesture is not fully registered, and scrollign stops abruptly
- Swiping back from the "List of Watches" screen also renders at very low FPS.
- I also entered a weird state while manipulating the app to report this issue, see the video linked below.

Is it because the app has a performance bug, or is it the performance Compose offers on WearOS?

I was considering making an app with WearOS Compose, but I'm a bit worried about the fact that using my app would be the opposite of enjoyable because of such poor performance that feels all but natural.

I made a video so you can see how it is, you'll see the difference between how smoothly my finger moves, versus how the ComposeAdvanced app reacts. Here's the Google Photos link: https://photos.app.goo.gl/weBQjfUk3NG9d7zL6 (I didn't attach it directly because it's over the 100MB limit, despite being only a bit more than one minute).

Have a great day,

Louis CAD

yschimke:
Are you running with a release build?

yschimke:
The compose team can best answer this in compose-wear in slack.  But it's known that there is a big difference for debug builds, but similar performance (views vs compose) with release builds.

If it's with a release build, then it's a bug.  I have a lot more complicated wear compose apps, without any of the issues you are describing.

yschimke:
To give one personal data point, I struggle to run debug builds on device, while they work fine in an emulator.  But release builds work smoothly on device.

LouisCAD:
I'm running it all with the debug buildType since the release buildType lacks the signing config in the sample.

What I showed is the default experience. If the performance is that poor on the device in debug mode, and if the only way around is building with minifyEnabled, maybe the default config should be edited?

yschimke:
Ooops, yep, good catch.

PR #239, opened 2022-03-13
Release builds for samples

yschimke:
Fixes https://github.com/android/wear-os-samples/issues/238#issuecomment-1065984919

### DIFF
--- a/.github/workflows/AlwaysOnKotlin.yml
+++ b/.github/workflows/AlwaysOnKotlin.yml
@@ -124,15 +124,28 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
-      - name: set up JDK 11
-        uses: actions/setup-java@v3
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
         with:
-          distribution: 'zulu'
           java-version: 11
 
-      - name: build release
-        run: ./gradlew assembleRelease
+      - name: Generate cache key
+        run: ./scripts/checksum.sh $SAMPLE_PATH checksum.txt
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches/modules-*
+            ~/.gradle/caches/jars-*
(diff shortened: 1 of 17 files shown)


LouisCAD:
Could you also add an `isDefault = true` to the release buildTypes of the Compose app so it's the default selected by Android Studio?

LouisCAD:
I'm also thinking that it might not be enough: When I tried the samples, I built them by running `.gradlew installDebug`.

I wonder if R8 shouldn't be enabled on debug builds too by default, albeit with obfuscation disabled.

Thinking a bit deeper, I guess it'd make sense to have a "Slow mode" ribbon, like debug Flutter apps have when `minifyEnabled` is left/set to false. That'd make people search for the non slow mode that requires minification to be turned on.

.github/workflows/AlwaysOnKotlin.yml
```diff
@@ -119,3 +119,36 @@ jobs:
...
+      - name: Generate cache key
+        run: ./scripts/checksum.sh $SAMPLE_PATH checksum.txt
+
+      - uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches/modules-*
+            ~/.gradle/caches/jars-*
+            ~/.gradle/caches/build-cache-*
+          key: gradle-${{ hashFiles('checksum.txt') }}
```
Goooler, line 144: Use `gradle/gradle-build-action@v2` ?

shumelchyk (approved):
Yuri, PR looks good, let's also handle suggestions from @LouisCAD

yschimke:
I think it's non controversial to change the default to release for any compose modules.  But I'll hold off on the Slow mode ribbon.

yschimke:
I'm not confident about the behaviour of unobfuscated debug builds for general debugging, so I'll hold off on that for now.

shumelchyk:
> so I'll hold off on that for now
I agree on this because Compose for WearOS is still in alpha version and we have planned improvements on the roadmap.

yschimke:
This breaks the previews.  Might add a comment in the code.  We can't have both unless we ship the Preview classes in release.

https://github.com/android/wear-os-samples/pull/247#discussion_r832465742

PR #241, opened 2022-03-14
Remove false char limits

LouisCAD:
The descriptions strings are not actually limited to 20 characters,
and it wasn't respected anyway.


yschimke:
cc @ADClarke to confirm.

ADClarke:
I'm not the best reviewer for this since I didn't author this watch face.  That said this is probably OK, the strings are long enough they probably wrap to multiple lines anyway.

garanj:
CHAR LIMIT=N annotations are generally used to help with internal Google translation processes, so I don't think they're hugely useful here. Agree removing them for clarity makes sense.

PR #243, opened 2022-03-16
Set @Preview options for full page previews

castedmo:
Starting with Android Studio Dolphin, showSystemUi option controls if
the device frame/clip should be applied.


yschimke (approved):
Thanks

PR #249, opened 2022-03-25
Update ComposeAdvanced to Wear Compose 1.0.0-alpha19 and add BaselineProfile

BowerSteve:
Also updated to Wear Compose 1.0.0-alpha19.


ComposeAdvanced/benchmark/src/main/java/com/example/android/wearable/composeadvanced/benchmark/BaselineProfile.kt
```diff
@@ -0,0 +1,114 @@
...
+    @Before
+    fun setUp() {
+        val instrumentation = InstrumentationRegistry.getInstrumentation()
+        device = UiDevice.getInstance(instrumentation)
+    }
+
+    @Test
+    fun profile() {
+        baselineRule.collectBaselineProfile(
+            packageName = PACKAGE_NAME,
```
yschimke, line 60: Having this demo is really useful.  We should screen record it and have it in a readme in the benchmark directory. Not suggesting now, but we should make this the wear benchmark sample.

ComposeAdvanced/build.gradle
```diff
@@ -24,6 +24,8 @@ buildscript {
 plugins {
     alias(libs.plugins.com.diffplug.spotless) apply(false)
     alias(libs.plugins.com.android.application) apply(false)
+    id 'com.android.test' version '7.1.1' apply false
+    id 'org.jetbrains.kotlin.android' version '1.6.10' apply false
```
yschimke, line 28:
```suggestion    
    alias(libs.plugins.com.android.test) apply(false)
    alias(libs.plugins.org.jetbrains.kotlin.android) apply(false)
```

PR #250, opened 2022-03-31
Add Macrobenchmark Startup test and baseline-prof.txt

BowerSteve:
Bug: 227564392


yschimke (approved):

yschimke:
@BowerSteve I suspect that androidMain isn't picked up.  This isn't a multiplatform project, so I'm not sure that androidMain is recognised in the path.

I think this should move to main.

Issue #246, opened 2022-03-22
Previews fail by default because we use release as default buildVariant


Digipom:
The starter code is missing this dependency:

```implementation libs.compose.ui.tooling```

Without it, I get this error:

```The following classes could not be found:
- androidx.compose.ui.tooling.ComposeViewAdapter (Fix Build Path, Edit XML, Create Class)```

yschimke:
Not missing for debug, but we changed the default to release builds.

PR #253, opened 2022-04-01
Fix Previews and revert default to debug 

yschimke:
Too many downsides to using the release as default, previews don't work and androidTest can't run.

Towards https://github.com/android/wear-os-samples/issues/246


shumelchyk:
I had some doubts when we made release builds the default, but decided that we should give it a try.
I think it will be useful to add a paragraph about performance in the Compose Advanced project's README to say that performance should be measured only on release builds with R8 and minify enabled.

yschimke:
I've put a paragraph here.  If we tweak it a bit, and then update other projects.

PR #255, opened 2022-04-01
User debug builds.

yschimke:


shumelchyk:
Same comment as in https://github.com/android/wear-os-samples/pull/253#issuecomment-1086208051

yschimke:
@shumelchyk I'm going to land now and in a week or so, update these samples readmes.

PR #261, opened 2022-04-05
AlwaysOn: Update gradle dependency versions

yschimke:


yschimke:
Failing on known problem tests.

shumelchyk (approved):
Any reasons why you've decided to split dependency updates into PRs for each project separately?

yschimke:
> Any reasons why you've decided to split dependency updates into PRs for each project separately?

They are different gradle projects, with different github builds, so some can pass and land while I fix others.

The automatic update command I'm using runs a single project at a time.

PR #263, opened 2022-04-05
Compose Advanced: Update dependencies

yschimke:


shumelchyk:
For this and other PRs for updating dependencies - do we want to keep one version catalog on a root-level for all Wear OS samples for automation purposes or it doesn't needed?

yschimke:
> For this and other PRs for updating dependencies - do we want to keep one version catalog on a root-level for all Wear OS samples for automation purposes or it doesn't needed?

I did previously, it doesn't work because the samples get imported to Android Studio as a single directory and must be self contained.

yschimke:
@shumelchyk Going to land anyway, please let me know if you have concerns.  But I'll be testing this further in a draft PR.

shumelchyk:
Ah I see, makes sense!

PR #269, opened 2022-04-14
Testing with Wear Compose snapshot

yschimke:
To land after wear-compose-alpha21 is available


yschimke:
This was a pretty hard fail until I bumped kotlin to 1.6.20 and AGP to 7.3.0-alpha.  Not sure which fixed it, and now it complains if I try to go back.

PR #311, opened 2022-04-20
Use common versions in catalog.

yschimke:


yschimke:
Landing, with a small risk of breaking to flush out issues.

PR #335, opened 2022-05-04
TimeText and View model cleanup

yschimke:
As I was updating view models, I pulled a thread.  Ideally this would be two PRs, happy to split if required. 

1) Fade away time text on list scroll using the viewmodel mechanism
2) Cleanup the other view models, based on the brief guidance from @astamato 

@astamato Including you because of the ViewModel changes. If it lands before you take a look, I can revisit with a follow up, so no rush.


ComposeAdvanced/gradle/libs.versions.toml
```diff
@@ -1,11 +1,12 @@
 [versions]
 android-gradle-plugin = "7.1.3"
-androidx-activity = "1.4.0"
-androidx-lifecycle = "2.4.1"
+androidx-activity = "1.5.0-beta01"
```
yschimke, line 3: I figured this should be safe since they are beta, should be final before wear-compose?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/watch/WatchDetailScreen.kt
```diff
@@ -26,43 +28,36 @@ import androidx.compose.foundation.layout.size
...
+import com.google.android.horologist.compose.navscaffold.ExperimentalComposeLayoutApi
+import com.google.android.horologist.compose.navscaffold.scrollableColumn
 
 /**
  * Displays the icon, title, and description of the watch model.
  */
 @Composable
 fun WatchDetailScreen(
-    id: Int,
+    watch: WatchModel?,
```
yschimke, line 53: @astamato was this what you were suggesting?

JohnNichol: This looks much better to me

JohnNichol (approved):

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/WearApp.kt
```diff
@@ -180,10 +196,33 @@ fun WearApp(
...
+                ) {
+                    val passedScrollType = it.arguments?.getSerializable(SCROLL_TYPE_NAV_ARGUMENT)
+
+                    check(
+                        passedScrollType == DestinationScrollType.SCALING_LAZY_COLUMN_SCROLLING
+                    ) {
+                        "Scroll type must be DestinationScrollType.SCALING_LAZY_COLUMN_SCROLLING"
+                    }
+
+                    val scrollViewModel: ScalingLazyListStateViewModel = viewModel(it)
```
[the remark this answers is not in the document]
yschimke: That's horologist :)

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/landing/LandingScreen.kt
```diff
@@ -56,18 +62,21 @@ import com.example.android.wearable.composeadvanced.presentation.ui.util.ReportF
...
 @Composable
 fun LandingScreen(
+    scalingLazyListState: ScalingLazyListState,
+    focusRequester: FocusRequester,
     onClickWatchList: () -> Unit,
     onClickDemoUserInputComponents: () -> Unit,
     onClickDemoMap: () -> Unit,
     onClickDialogs: () -> Unit,
     proceedingTimeTextEnabled: Boolean,
     onClickProceedingTimeText: (Boolean) -> Unit,
```
astamato, line 72: Consider adding a modifier to all composable functions

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/map/MapScreen.kt
```diff
@@ -37,11 +38,9 @@ import com.google.maps.android.compose.rememberCameraPositionState
  * and put it in local.properties with key `mapsApiKey`.
  */
 @Composable
-fun MapScreen(
-    modifier: Modifier = Modifier,
```
astamato, line 41: Why was the default modifier removed?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/landing/LandingScreen.kt
yschimke: https://github.com/android/wear-os-samples/pull/351

astamato:
Instead of saving your state in ScalingLazyListStateViewModel, can you achieve the same with rememberSaveable? which also helps retain state through process death

yschimke:
@astamato The critical function of the ScalingLazyListStateViewModel is actually a coordination point across the top level activity Composable, and the changing navigation destination Composables. The top level one accesses this through

```        // Observes the current back stack entry to pull information and determine if the screen
        // is scrollable and the scrollable state.
        //
        // The main reason the state for any scrollable screen is hoisted to this level is so the
        // Scaffold can properly place the position indicator (also known as the scroll indicator).
        //
        // We save the above scrollable states in the SavedStateHandle and retrieve them
        // when needed from custom view models (see ScrollingViewModels class).
        //
        // Screens with scrollable content:
        //  1. The watch list screen uses ScalingLazyColumn (backed by ScalingLazyListState)
        //  2. The watch detail screens uses Column with scrolling enabled (backed by ScrollState).
        //
        // We also use these scrolling states for various other things (like hiding the time
        // when the user is scrolling and only showing the vignette when the screen is
        // scrollable).
        //
        // Remember, mobile guidelines specify that if you back navigate out of a screen and then
        // later navigate into it again, it should be in its initial scroll state (not the last
        // scroll location it was in before you backed out).
        val currentBackStackEntry by swipeDismissableNavController.currentBackStackEntryAsState()
...
                        DestinationScrollType.SCALING_LAZY_COLUMN_SCROLLING -> {
                            val scrollViewModel: ScalingLazyListStateViewModel =
                                viewModel(currentBackStackEntry!!)
```

Also I believe these are saved here

https://github.com/android/wear-os-samples/blob/6c707cec5e9202b04c6fd2f0ff6f322598a4d746/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/ScrollingViewModels.kt#L27

PR #339, opened 2022-05-04
Show Map as a separate Activity

yschimke:
image

### DIFF
--- a/ComposeAdvanced/app/src/main/AndroidManifest.xml
+++ b/ComposeAdvanced/app/src/main/AndroidManifest.xml
@@ -53,7 +53,7 @@
 
         <activity
             android:name=".presentation.ui.map.MapActivity"
-            android:exported="true"
+            android:exported="false"
             android:label="@string/map_activity_name">
         </activity>
 
--- a/ComposeAdvanced/app/src/main/AndroidManifest.xml
+++ b/ComposeAdvanced/app/src/main/AndroidManifest.xml
@@ -40,8 +40,7 @@
 
         <activity
             android:name=".presentation.MainActivity"
-            android:exported="true"
-            android:label="@string/app_name">
+            android:exported="true">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />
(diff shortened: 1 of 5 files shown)


yschimke:
https://user-images.githubusercontent.com/231923/166715715-3b129329-218e-4737-bd3b-71f1ad6b3bd3.mp4

ComposeAdvanced/app/src/main/AndroidManifest.xml
```diff
@@ -52,6 +51,12 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name=".presentation.ui.map.MapActivity"
+            android:exported="true"
```
alexvanyo, line 56: This doesn't probably need to be exported, if it's only being launched from within the app.

yschimke: Is there any clean navigation way to launch an intent? I couldn't find one and I think intent in the deeplinkoptions is just a naming option to a route.

alexvanyo: Does https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation/navigation-runtime/src/main/java/androidx/navigation/ActivityNavigatorDestinationBuilder.kt;l=49;drc=dcb2c081d6802481afd60131f935bbcb9b0663dc work in Compose? I haven't actually tried.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/map/MapScreen.kt
```diff
@@ -51,18 +56,27 @@ fun MapScreen() {
...
+                )
+            }
+            if (onClose != null) {
+                Button(
+                    modifier = Modifier
+                        .align(Alignment.BottomCenter)
+                        .size(ButtonDefaults.SmallButtonSize),
+                    onClick = onClose,
+                ) {
+                    Text(text = "X")
```
BowerSteve, line 77: How about using a CompactChip here, which would take up less of the screen? Like the AM/PM chip on the TimePickerWith12HourClock demo:
`            CompactChip(
                onClick = {},
                modifier = Modifier.size(width = 50.dp, height = 24.dp),
                label = {
                    Row(
                        modifier = Modifier.fillMaxWidth(),
                        horizontalArrangement = Arrangement.Center
                    ) {
                        Text(
                            text = "X",
                            color = MaterialTheme.colors.onSecondary,
                            style = MaterialTheme.typography.button,
                        )
                    }
                },
                colors = ChipDefaults.chipColors(backgroundColor = MaterialTheme.colors.secondary),
                contentPadding = PaddingValues(vertical = 0.dp),
            )
`

BowerSteve (approved):
Looks good - would be cool to tidy up the close button as per comment.

PR #340, opened 2022-05-04
Add progress indicator examples to compose advanced

JohnNichol:
Add an example of a spinner and a full screen progress indicator to the sample.
Add a new type of destination scroll type for time text only.
Fix up out of date comment in UserInputComponentsScreen.kt


yschimke (approved):
Screenshots would help reviewers get an idea of the feature.

PR #344, opened 2022-05-05
Put a gradient under the Watch identifier

JohnNichol:
The Round Watch/Square Watch text at the bottom of the main screen is hard to read as it is purple on purple a lot of the time. So adding a gradient behind which also allows us to show how to do modifier/gradients in the curved world


JohnNichol:
Screenshot_20220505_140158
Screenshot_20220505_140223

PR #346, opened 2022-05-05
Refactor Landingscreen

JohnNichol:
Allow LandingScreen to take a map of menu item names to click handlers - before adding more items to the list I wanted to do a refactor and also check that folks are happy with this approach as I think we can apply it to some of the other menus as well.


ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/WearApp.kt
```diff
@@ -228,30 +227,40 @@ fun WearApp(
 
                     val focusRequester = remember { FocusRequester() }
 
-                    val context = LocalContext.current
+                    val menuItemNameToClickHandlerMap = sortedMapOf(
```
yschimke, line 230: sortedMap seems like an unlikely realistic choice.  What about LinkedHashMap, so insertion order is respected?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/WearApp.kt
```diff
@@ -505,6 +514,13 @@ fun WearApp(
     }
 }
 
+@Composable
+private fun menuNameAndCallback(
```
yschimke, line 518: TIOLI: Why not just a simple data class with resource id, and screen.  Leave navigate and stringResource to the LandingScreen?

yschimke (approved):
Nothing blocking, just some optional suggestions.

After completing the change, LGTM.

PR #351, opened 2022-05-07
Modifiers

yschimke:
Add in Modifier.

Follow best practices with Modifier as the first optional parameter.


astamato (approved):
Lgtm, TY

PR #354, opened 2022-05-09
Add dynamic/selectable theme support to the Advanced Compose add.

JohnNichol:
Give the user access to 4 different theme alternatives and a way to select them in the app

Screenshot_20220509_185158

Screenshot_20220509_185116
.

### DIFF
--- a/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/WearApp.kt
+++ b/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/WearApp.kt
@@ -64,6 +64,8 @@ import com.example.android.wearable.composeadvanced.presentation.navigation.SCRO
 import com.example.android.wearable.composeadvanced.presentation.navigation.Screen
 import com.example.android.wearable.composeadvanced.presentation.navigation.WATCH_ID_NAV_ARGUMENT
 import com.example.android.wearable.composeadvanced.presentation.theme.WearAppTheme
+import com.example.android.wearable.composeadvanced.presentation.theme.initialThemeValues
+import com.example.android.wearable.composeadvanced.presentation.theme.themeValues
 import com.example.android.wearable.composeadvanced.presentation.ui.ScalingLazyListStateViewModel
 import com.example.android.wearable.composeadvanced.presentation.ui.ScrollStateViewModel
 import com.example.android.wearable.composeadvanced.presentation.ui.dialog.Dialogs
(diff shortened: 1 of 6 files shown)


ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/theme/ThemeScreen.kt
```diff
@@ -0,0 +1,75 @@
...
+    ) {
+        item {
+            ListHeader {
+                Text("Color Schemes")
+            }
+        }
+        for (listItem in availableThemes) {
+            val checked = listItem.colors == currentlySelectedColors
+            item {
+                ToggleChip(
```
yschimke, line 54: What about

```suggestion
MaterialTheme(colors = listItem.colors) {
                ToggleChip(
```

For a bit of pizzazz.

yschimke (approved):

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/WearApp.kt
```diff
@@ -503,6 +512,29 @@ fun WearApp(
...
+                ) { it ->
+                    val scalingLazyListState = scalingLazyListState(it)
+                    val focusRequester = remember { FocusRequester() }
+
+                    ThemeScreen(
+                        scalingLazyListState = scalingLazyListState,
+                        focusRequester = focusRequester,
+                        currentlySelectedColors = themeColors,
+                        availableThemes = themeValues
+                    ) { colors -> themeColors = colors }
```
yschimke, line 534: Curious: is there a agreed style for event handlers as last method param? named vs trailing lambda?

shumelchyk (approved):
Nice one 👍

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/theme/ThemeScreen.kt
JohnNichol: As per offline discussion going to leave this as is. The override of selection control icon color is being done as the normal selection control color is the "secondary" color - so if you select the blue theme you get a green selection control color - which in this very specific use case is a potentially confusing given the colors are changing as you select items in the list.

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/WearApp.kt
JohnNichol: I think this falls under the general callback as last property to allow for lambdas unless it causes confusion. But I agree with compose where we will often have a content lambda at the end we will often end up naming callbacks. I think the basic premise should be considering have last param as something lambda'erable if you can.

PR #357, opened 2022-05-11
Only run AlwaysOnKotlin on 28

yschimke:
26 and 28 are working, but don't meet test requirements.


yschimke:
Going to land to fix build, feel free to ask me to revert, but if so we should plan for a fix.

PR #358, opened 2022-05-11
Fix Preview bitmap not updating 

lancelinksp:
Preview bitmap was not updating while changing settings. Added yield() to fix problem


yschimke:
LGTM - thanks for verifying the suggested fix.

lancelinksp:
@yschimke The CLA was previously completed and I performed the rescan. Is there a possibility that I missed something else? Another step possibly?

yschimke (approved):

alexvanyo (approved):
Looks good to me, thanks for the fix!

PR #397, opened 2022-06-02
Add rotary support to the Tiles Kotlin screen

ataulm:


WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/MainActivity.kt
```diff
@@ -36,7 +36,7 @@ class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContent {
-            SampleTilesList(this, sampleTiles)
+            TilesKotlinScreen(context = this, sampleTiles = sampleTiles)
```
ataulm, line 39: Is this the correct naming convention for top level screens?

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/TilesKotlinScreen.kt
```diff
@@ -21,20 +21,33 @@ import androidx.annotation.StringRes
...
 import androidx.wear.tiles.TileService
+import com.google.android.horologist.compose.navscaffold.ExperimentalHorologistComposeLayoutApi
+import com.google.android.horologist.compose.navscaffold.scrollableColumn
+import com.google.android.horologist.compose.pager.FocusOnResume
 
+@OptIn(ExperimentalHorologistComposeLayoutApi::class)
 @Composable
-internal fun SampleTilesList(context: Context, sampleTiles: Map<Int, Class<out TileService>>) {
+internal fun TilesKotlinScreen(
+    modifier: Modifier = Modifier,
```
yschimke, line 40: nit: Modifier should be first optional parameter

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/TilesKotlinScreen.kt
```diff
@@ -21,20 +21,33 @@ import androidx.annotation.StringRes
...
+    context: Context,
+    sampleTiles: Map<Int, Class<out TileService>>,
+    focusRequester: FocusRequester = remember { FocusRequester() }
+) {
     TilesKotlinTheme {
         val listState = rememberScalingLazyListState()
         ScalingLazyColumn(
-            modifier = Modifier.fillMaxSize(),
+            modifier = modifier
+                .fillMaxSize()
```
yschimke, line 49: I'd probably assume the caller is deciding this part, possibly the parent is setting `fillMaxSize(0.9)` and you are overriding.

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/TilesKotlinScreen.kt
```diff
@@ -49,6 +62,7 @@ internal fun SampleTilesList(context: Context, sampleTiles: Map<Int, Class<out T
             }
         }
     }
+    FocusOnResume(focusRequester = focusRequester)
```
yschimke, line 65: nit: I'd probably assume who ever is creating this knows the timing of focus requests.  But focusRequester is a param with a default value, so it isn't clear here.  Maybe either move FocusOnResume up to MainActivity and make focusRequester a mandatory param.  Or the opposite, remove as a param and create in the body of this method.

yschimke (approved):
Nice, enjoy your long weekend!

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/TilesKotlinScreen.kt
ataulm, line 65: It would help me if you could comment on each statement with confirmation/clarifications 🙏 

1. a composable (like this screen) can/should request focus for descendants, like the SLC. It's not appropriate for SLC to request focus on itself
2. here, the screen is not requesting focus immediately, it's using lifecycle callbacks to _then_ request focus. Are there cases where requesting focus immediately (i.e. `focusRequester.requestFocus()`) makes sense?
3. We're passing `focusRequester` to the `scrollableColumn` modifier above. This is not used to request focus. It's so that we (the scrollableColumn modifier) can be notified about changes to focus.

`#3` is the most difficult one to wrap my head around, presumably because I don't understand how Compose is working with these events.

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/TilesKotlinScreen.kt
ataulm: Is this the case for composables that are intended to be the root screen, or is "root screen" an assumption that doesn't always hold?

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/TilesKotlinScreen.kt
yschimke: 1) Yes. An individual component can't know that it should be the requested item to start with. There can be multiple on a single screen, think of a DatePicker.  Or there can be multiple screens composed at same time.  Think HorizontalPager, SwipeToDismissBox or SwipeDismissableNavHost.

2) I'm not 100% definitive on the definitions. My understanding is focusable, makes the item focusable through general focus navigation, e.g. dtab, or textfields. focusRequester is a controller that lets you do this programmatically.

See https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/samples/src/main/java/androidx/compose/ui/samples/FocusSamples.kt;l=103?q=RequestFocusSample

It's hard to think about cases where immediately requesting focus makes sense, since you don't know if you are composed before you ever come on screen if at all.

3) I think answered in 2. I don't think the focusRequester is about getting notified of focus, I think it's about programmatically requesting focus in a UI toolkit that doesn't have 30 fields of views that you can call methods on.

yschimke:
The general problem is that for a sample it's always convenient to handle the focus close to the component (SLC). But in an app with multiple screens, it breaks down and scroll just doesn't work as expected, or scrolls something off screen. You can't make something like a DatePicker that actively manages focus internally, without assuming a lot about how you are positioned.

Maybe your date and time pickers are shown in 30% sized previews on screen, and then you animate one of them to full size when you click on it.  In that case two are on screen, should either receive the scroll events while minimised, and it's not the composition (LaunchedEffect) that is significant, but something that is only obvious to the parent container.

ataulm:
>Maybe your date and time pickers are shown in 30% sized previews on screen, and then you animate one of them to full size when you click on it. In that case two are on screen, should either receive the scroll events while minimised, and it's not the composition (LaunchedEffect) that is significant, but something that is only obvious to the parent container.

In this case, would the parent "decide" by choosing which descendant to apply the `scrollableColumn` (or equivalent) modifier?

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/TilesKotlinScreen.kt
ataulm: nice, I'll play with that sample 🙌  thank you

yschimke:
> > Maybe your date and time pickers are shown in 30% sized previews on screen, and then you animate one of them to full size when you click on it. In that case two are on screen, should either receive the scroll events while minimised, and it's not the composition (LaunchedEffect) that is significant, but something that is only obvious to the parent container.
> 
> In this case, would the parent "decide" by choosing which descendant to apply the `scrollableColumn` (or equivalent) modifier?

I assume by having two focusRequesters, and calling requestFocus when each is maximised. Both would be scrollableColumn from the start, but with two different instances of FocusRequester.

Issue #402, opened 2022-06-08
Create Tile showcasing multiple timeline entries


ataulm:
The 3 tiles we have today only show a single entry: fitness, messaging, playNext.

Create a calendar event tile with multiple entries to showcase its usage.

yschimke:
Consider using one of the current ones, maybe it's enough to fallback from fitness tile with an active activity, to a blank one to launch an activity.  Sort of like a TTL, if app is running it keeps changing the tile.  Once it stops the activity clears after some TTL.

PR #403, opened 2022-06-08
Remove onTerminate in TilesApplication

### LINKED ISSUE #401: Remove TilesApplication.onTerminate
Not called in practice, and maybe AppContainer should avoid an application owned CoroutineScope.

ataulm:
- Fixes #401 
- Adds `singleEntryTimeline` to reduce boilerplate. Added new issue #402 because we don't demo multiple entry timelines
- Removes the updates mechanism since it's unused


WearTilesKotlin/app/src/main/java/com/example/wear/tiles/SingleEntryTimeline.kt
```diff
@@ -0,0 +1,22 @@
+package com.example.wear.tiles
+
+import androidx.wear.tiles.LayoutElementBuilders
+import androidx.wear.tiles.LayoutElementBuilders.LayoutElement
+import androidx.wear.tiles.TimelineBuilders
+
+/**
+ * Provides a [TimelineBuilders.Timeline] which has only one entry.
+ */
+internal fun singleEntryTimeline(rootLayout: LayoutElement): TimelineBuilders.Timeline {
```
ataulm, line 10: potential candidate for horologist-tiles?

ataulm: is this a DSL? 🤔

ataulm: + Do you mean in samples or Horologist?

yschimke:
Thanks!

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/SingleEntryTimeline.kt
ataulm: my mistake - I had the `{}` in MessagesRenderer but forgot to update that one bit!

PR #445, opened 2022-06-30
Remove desugar

yschimke:
Seemingly not needed.


yschimke:
Still needed according to https://developer.android.com/studio/write/java8-support

PR #450, opened 2022-07-10
Add Golden Tiles samples

ataulm:
Screen Shot 2022-07-11 at 05 35 22

This PR adds implementations for each of the Golden Tiles from the [Wear Tiles Design Kit](https://developer.android.com/training/wearables/design/tiles-design-system)
 
- Uses a snapshot of androidx tiles material components because of changes that aren't scheduled for new alphas
- `ComponentAudit.kt` should only show individual components (e.g. Chip). Layouts will all be covered by `GoldenTilesPreviews.kt`
- I've raised bugs for the ones for which I couldn't match the designs
- I've left two unimplemented: Sleep Tracker and the graph in Heart Rate; they're not showcasing Tile Material Components so there's no rush on these.

image

image

image

image

image


WearTilesKotlin/app/src/main/java/com/example/wear/tiles/golden/Workout.kt
```diff
@@ -0,0 +1,111 @@
...
+                    .addButtonContent(
+                        Button.Builder(context, button3Clickable).setIconContent(BUTTON_3_ICON_ID)
+                            .build()
+                    )
+                    .build()
+            )
+            .setPrimaryChipContent(
+                CompactChip.Builder(context, "More", chipClickable, deviceParameters)
+                    .setChipColors(
+                        ChipColors(
```
yschimke, line 71: bordering on the don't review for content, but would it be helpful to implement these designs using themes using Colors? then using primary or secondary ChipColors factories?

yschimke:
Direction looks, great.  Would you consider using the Horologist TileRenderer abstraction, should give a nice clean separation for the data of each tile.  But can understand why you may not.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/golden/Workout.kt
ataulm, line 71: I know we've got variants of this where it selects primary/surface colors from a pseudo-theme object, but I haven't seen the utility yet.

Tiles are often (always?) on dark themes, and they're so small and each app would have so few of them, it seems easier to specify colors explicitly.

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/golden/GoldenTilesPreviews.kt
```diff
@@ -0,0 +1,83 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.platform.LocalContext
+import com.example.wear.tiles.R
+import com.example.wear.tiles.emptyClickable
+import com.example.wear.tiles.tools.WearSmallRoundDevicePreview
+import com.google.android.horologist.compose.tools.LayoutRootPreview
+import com.google.android.horologist.compose.tools.buildDeviceParameters
+import com.google.android.horologist.tiles.images.drawableResToImageResource
+
+@WearSmallRoundDevicePreview
```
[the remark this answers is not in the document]
yschimke: Because https://issuetracker.google.com/issues/233511976

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/golden/Workout.kt
yschimke: https://github.com/yschimke/rememberwear/blob/main/wear/src/main/kotlin/com/google/wear/soyted/tile/RememberWearTileRenderer.kt#L116

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/golden/GoldenTilesPreviews.kt
```diff
@@ -0,0 +1,83 @@
...
+            weekSummary = "1 run this week",
+            button1Clickable = emptyClickable,
+            button2Clickable = emptyClickable,
+            button3Clickable = emptyClickable,
+            chipClickable = emptyClickable,
+        )
+    ) {
+        addIdToImageMapping(
+            Workout.BUTTON_1_ICON_ID,
+            drawableResToImageResource(R.drawable.ic_run_24)
```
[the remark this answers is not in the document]
yschimke: Nice idea.

ataulm:
> Direction looks, great. Would you consider using the Horologist TileRenderer abstraction, should give a nice clean separation for the data of each tile. But can understand why you may not.

Will be able to see when I move onto updating the samples themselves :+1:

---

In my mind, the samples should be used to demonstrate some of the common or complex use cases. at the moment, we're using them to showcase 3 layouts which won't be necessary when this is merged, since we'll be demonstrating all the Golden Tile layouts without the overhead of having a full tile for each.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/golden/Workout.kt
ataulm: I added a comment for this elsewhere. I'm not sure themes for tiles is worth it - as I was implementing all of these, I found it easier to specify colors explicitly.

In a real project, I imagine it'll be even less worth it:

- a real project won't have many tiles
- themes aren't directly reusable between Compose/this
- even if they were reusable, it's not guaranteed that we'd want the same combination of colors
- tiles (/wear?) is always dark mode and doesn't support multi-theming

I'm not sure what it provides.

WearTilesKotlin/app/src/debug/java/com/example/wear/tiles/golden/GoldenTilesPreviews.kt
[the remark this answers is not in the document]
ataulm: +1'd!

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/golden/Workout.kt
ataulm: discussion merged, go here -> https://github.com/android/wear-os-samples/pull/450#discussion_r917425999

ataulm: The link you provided here to an example project does help show some utility - https://github.com/yschimke/rememberwear/blob/main/wear/src/main/kotlin/com/google/wear/soyted/tile/RememberWearTileRenderer.kt#L116

---

There's an example of a theme being created in `ComponentAudit.kt` which we can leave as-is / improve. 

I think the Golden Tiles don't benefit from themes because it'll only serve to showcase the overhead without any of the benefit because it's not being shared anywhere (vs. your RememberTheMilk project which was sharing it with the main app).

WearTilesKotlin/app/build.gradle
```diff
@@ -63,29 +63,25 @@ android {
 }
 
 dependencies {
-    implementation libs.horologist.compose.layout
```
ataulm, line 66: just alphabetized

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/golden/Workout.kt
yschimke: IF we want people to be able to take these tiles and copy and paste them, I think it's useful. Or maybe a library of reference tile designs, "templates" that we provide.  But your call.

ataulm: Let's leave it for now and we'll revisit if there's feedback. I think these samples should facilitate understanding and we shouldn't necessarily be encouraging copy and pasting.

PR #490, opened 2022-08-02
Remove redundant Tiles samples + preview activity

ataulm:
The 3 tile samples we had showcased different layouts but these are mostly redundant now that we have the Golden Tiles previews which showcase even more layouts.

This PR removes a bunch of things:

- 2 of the 3 samples (fitness/play next song) and keeps the Messaging sample since it's the most advanced of the 3. a follow up patch will update the Messaging sample UI to use Tiles Material Components.
- TilesPreviewActivity since we have compose Previews in AS, and the tiles surface launcher in AS Dolphin
- Unused files and dependencies.


WearTilesKotlin/app/build.gradle
```diff
@@ -63,16 +63,11 @@ android {
 }
 
 dependencies {
-    implementation libs.androidx.activity.compose
+    implementation libs.androidx.lifecycle.runtime.ktx
```
ataulm, line 66: This was a transitive dependency of the activity compose library but since we don't have any activities any more, I added the lifecycle dependency directly

WearTilesKotlin/app/src/debug/AndroidManifest.xml
```diff
@@ -17,20 +17,4 @@
     package="com.example.wear.tiles">
 
     <uses-feature android:name="android.hardware.type.watch" />
-
-    <application>
-
-        <activity
-            android:name=".MainActivity"
```
yschimke, line 24: Q: Is it useful to have an Activity in case you want to launch something on a click?

yschimke (approved):
LGTM

WearTilesKotlin/app/src/debug/AndroidManifest.xml
ataulm: yep but let's add it back when we do that (maybe in the next pr + 1)

ataulm: good call out - atm we just have the no-op clicks and it's worth showing how to handle it

Issue #496, opened 2022-08-04
Unclear what to do when ResourcesRequest doesn't have resource ids


ataulm:
`onResourcesRequest(requestParams: ResourcesRequest)` 

the `requestParams.resourceIds` might be empty. When does this happen? What's the expected behavior from a tile developer in this case?

yschimke:
Documented here https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/tiles/tiles/src/main/java/androidx/wear/tiles/RequestBuilders.java;l=127?q=ResourcesRequest

But it's hard to implement for say a Twitter tile, without remembering which Tiles you returned.

ataulm:
The documentation says to return all the resources for a given version if the list is empty (I wonder why the list can't just be exhaustive to save developers from having to do this check). It doesn't say why/when it will happen and I wonder if understanding this would be helpful to verify that we (as tile devs) are doing the correct thing for our particular cases.

Let's revisit this holistically inc the following comment:

>SingleTileLayoutRenderer is hiding the complexity of versions and effect on resources. I wonder whether we should expose this to override in SingleTileLayoutRenderer, or if not, put a quick doc comment here to explain that images will be cached permanently, and ids should have a timestamp if we want them to evolve, or we should change the id to flush all...

https://github.com/android/wear-os-samples/pull/497#discussion_r938692205

It looks like resource versioning and resource IDs are related, and I think a good outcome for this issue would be a blog post explaining them, with examples. Perhaps in the same post as #495 if it makes sense.

yschimke:
I think a good approach might be to end with different scenarios that would be good reasons for the following versioning/resource strategies

- Simple app that returns static images that wouldn't change except for a release.  Fixed version, and simple ids.
- App that returns ids that are not unique, so bumps versions to request refetching of images.
- App that returns a constant value, but expects new images (say a new contact added) or updates existing, by appending a timestamp to the id.

yschimke:
Closing - the TileService never actually receives resource ids, so must generate all images on every request.

Issue #495, opened 2022-08-04
Handling errors on image loading for Tiles


ataulm:
onResourcesRequest() is where we fetch bitmaps from the network, convert to an image resource, and then return it as a resource bundle when requested.

What happens if something fails in that pipeline? What is the correct behavior (for tile developers)?

yschimke:
We need to replace with a permanent replacement image, and refresh resources another way.  Or fetch images while loading the tile, so we know what resource id to put.

ithinkihaveacat:
Closing as WONTFIX. (There is however now some rudimentary error handling that displays an offline image if loading fails.)

PR #497, opened 2022-08-05
Update tiles sample for 1.1.0

ataulm:
Before | After
---|---
image | tile_messaging

- Updated Messaging Tile to match one of the Golden Layouts (Timer)
- Updated the README to add references to Tiles Material Components and Horologist
- Add copyrights to XML files (and corresponding Spotless setup)
- Removed unused resources, moved Golden Tiles back to debug (since we're not using it in the sample, and they can be previewed with the composable annotation in the IDE)
- Extracted `messagingTileLayout` from `MessagingTileRenderer`


WearTilesKotlin/README.md
```diff
@@ -1,28 +1,29 @@
...
-convenient, designed to provide users with what they need in the moment, which gives users control
-over what information and actions they want to see.
+[Tiles](https://d.android.com/training/wearables/tiles) provide easy access to the information and
+actions users need in order to get things done. With a simple swipe from the watch face, a user can
+find out the latest forecast or start a timer.
 
-This repository contains a set of sample implementations of Tiles, demonstrating various use cases.
-Each Tile extends TileProviderService and implements onTileRequest and onResourcesRequest.
+This repository contains a sample tile built using the [Tiles Material Components](https://developer.android.com/reference/androidx/wear/tiles/material/package-summary)
+library (for UI) and [Horologist Tiles](https://github.com/google/horologist#-tiles) (preview
```
ataulm, line 11: Is this ok for Horologist?

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingRepo.kt
```diff
@@ -40,15 +40,15 @@ class MessagingRepo(private val context: Context) {
     suspend fun updateContacts(contacts: List<Contact>) {
         context.dataStore.edit {
             it.clear()
-            knownContacts.forEachIndexed { index, contact ->
+            contacts.forEachIndexed { index, contact ->
```
ataulm, line 43: I think a bug. No change in behavior since `knownContacts` is what gets passed to this function from the outside, but for correctness, this function doesn't know about the `knownContacts` property.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingRepo.kt
```diff
@@ -60,10 +60,10 @@ class MessagingRepo(private val context: Context) {
                 id = 1, initials = "AC", name = "Ali C", avatarUrl = "$avatarPath/ali.png"
             ),
             Contact(
-                id = 2, initials = "FS", name = "Felipe S", avatarUrl = null
+                id = 2, initials = "TB", name = "Taylor B", avatarUrl = "$avatarPath/taylor.jpg"
             ),
             Contact(
-                id = 3, initials = "TB", name = "Taylor B", avatarUrl = "$avatarPath/taylor.jpg"
+                id = 3, initials = "FS", name = "Felipe S", avatarUrl = null
```
ataulm, line 66: Just swapped these around to keep the ones with images in the same location.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileTheme.kt
```diff
@@ -13,22 +13,17 @@
...
-
-    override fun onCreate() {
-        super.onCreate()
-
-        appContainer = AppContainer(this)
-    }
-
-    // Allows Coil ImageLoader singleton to work from anywhere
-    override fun newImageLoader(): ImageLoader = appContainer.imageLoader
+object MessagingTileTheme {
```
ataulm, line 20: how's this?

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileLayout.kt
```diff
@@ -0,0 +1,177 @@
...
+            .addButtonContent(searchLayout(context, emptyClickable))
+            .build()
+    ).setPrimaryChipContent(
+        CompactChip.Builder(
+            context,
+            context.getString(R.string.tile_messaging_create_new),
+            emptyClickable,
+            deviceParameters
+        )
+            .setChipColors(ChipColors.primaryChipColors(MessagingTileTheme.colors))
```
ataulm, line 68: this ok?

yschimke: Yeah, I think it's the right approach for samples, unless we want to advocate for dynamically changing themes.  Maybe for media, we should show that when artwork changes?

yschimke: Tangential: related to the repetition of context, context.getString(), deviceParameters, MessagingTileTheme.colors. This was something that I wanted to improve by having a Renderer type that could simplify these repetitions. https://github.com/google/horologist/blob/ab8ae486d43b55e46423979d9ad2aacf382d323c/tiles/src/main/java/com/google/android/horologist/tiles/render/SingleTileLayoutRenderer.kt

But not for here, maybe something we can review these and get a nice streamlined API.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileRenderer.kt
```diff
@@ -51,223 +31,34 @@ class MessagingTileRenderer(context: Context) :
...
     ): LayoutElement {
-        return tileLayout(state, deviceParameters)
+        return messagingTileLayout(state, context, deviceParameters)
     }
 
+    /**
+     * If we want to display an image in our layout, we have to add the image resource to the
+     * [Resources.Builder].
+     */
     override fun Resources.Builder.produceRequestedResources(
```
yschimke, line 41: TIOLI: SingleTileLayoutRenderer is hiding the complexity of versions and effect on resources. I wonder whether we should expose this to override in SingleTileLayoutRenderer, or if not, put a quick doc comment here to explain that images will be cached permanently, and ids should have a timestamp if we want them to evolve, or we should change the id to flush all...

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileRenderer.kt
```diff
@@ -51,223 +31,34 @@ class MessagingTileRenderer(context: Context) :
...
         resourceResults: Map<Contact, Bitmap>,
         deviceParameters: DeviceParametersBuilders.DeviceParameters,
         resourceIds: MutableList<String>
     ) {
         if (resourceIds.isEmpty() || resourceIds.contains(ID_IC_SEARCH)) {
```
yschimke, line 46: Should explain this logic of the resource ids here?

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileRenderer.kt
yschimke: Another thing that SingleTileLayoutRenderer hides is freshnessIntervalMillis, is it worth overriding in this class to document, even if we set it to 0, and say that updates are coming from externally pushed? Or set it to 5 minutes?

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -35,77 +35,87 @@ import kotlinx.coroutines.flow.map
...
                 initialValue = null
             )
     }
 
-    private fun buildState(contacts: List<Contact>): MessagingTileState {
-        return MessagingTileState(contacts)
-    }
-
+    /**
+     * Read the latest data, delegating to the [MessagingTileRenderer] which creates a layout to
```
yschimke, line 69: Is it useful to do a brief explanation of StateFlow and why it's used here?

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -35,77 +35,87 @@ import kotlinx.coroutines.flow.map
...
-    private suspend fun readTileState(): MessagingTileState {
+    /**
+     * Reads the latest state from the flow, and updates the data if there isn't any.
+     */
+    private suspend fun latestTileState(): MessagingTileState {
         var tileState = tileStateFlow.filterNotNull().first()
 
         if (tileState.contacts.isEmpty()) {
-            updateContacts()
+            refreshData()
```
yschimke, line 84: TIOLI: As a sample, maybe this deserves a comment that we only have this here because the demo doesn't have a real backend? In a real app, app, complications, tiles would all share this common data source, and data refreshes wouldn't be triggered by a tile being installed.

yschimke: Actually saw the comment below for refreshData, maybe could slightly expand that and refer to that from here.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileTheme.kt
```diff
@@ -13,22 +13,17 @@
...
+    private val purple = android.graphics.Color.parseColor("#C58AF9")
```
yschimke, line 21: Why parse color over Color(0xFFC58AF9) ? Do we have style guides on this?

yschimke (approved):
Looks great

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -35,77 +35,87 @@ import kotlinx.coroutines.flow.map
...
+     * For this sample, we're updating the repository with fake data
+     * ([MessagingRepo.knownContacts]).
+     */
+    private suspend fun refreshData() {
         repo.updateContacts(MessagingRepo.knownContacts)
     }
 
+    /**
+     * Downloads bitmaps from the network and passes them to [MessagingTileRenderer] to add as
+     * image resources (alongside any local resources).
```
yschimke, line 103: Should we talk about what happens if image loading fails?  How to mitigate? e.g. ensure you are using the same ImageLoader and cache that the main app uses, avoid hitting the network for these requests because network may not be available?

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileLayout.kt
ataulm: >unless we want to advocate for dynamically changing themes. Maybe for media, we should show that when artwork changes?

yeah let's see how it looks in practice

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileRenderer.kt
ataulm: For versioning, added a note here to revisit this https://github.com/android/wear-os-samples/issues/496

re: the freshness thing - in this sample, I'm not sure that highlighting it will make anything clearer because we never update the tile externally and setting to a non-zero value (>0 means system will trigger refreshes) doesn't help either because the data never changes.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
ataulm: I would leave it for now - I think that it will make more sense if the data store is externally updated (e.g. from a configuration overlay)? At which point we could spend more time leaning into this.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileTheme.kt
ataulm: nah I was going to ask you about this but forgot. Will update. (We don't have style guides on it afaik)

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
ataulm: We can come back and update this when we have a fallback strategy (#495), but I don't think it's worth adding that info here because it's not easily demonstrated by the sample; should be in a post/DAC.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
ataulm: What if the tile is the first thing that's added (before the overlay is opened)? In that case, I would expect the tile to be responsible for fetching the data initially.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileRenderer.kt
```diff
@@ -51,223 +31,34 @@ class MessagingTileRenderer(context: Context) :
...
+                /* id = */ ID_IC_SEARCH,
+                /* image = */ drawableResToImageResource(R.drawable.ic_search_24)
             )
         }
 
-        // Add the scaled & cropped avatar images
         resourceResults.forEach { (contact, bitmap) ->
-            val imageResource = bitmapToImageResource(bitmap)
-            // Add each created image resource to the list
             addIdToImageMapping(
```
yschimke, line 54: From the chat, I think you are right, we should check these before returning.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileTheme.kt
yschimke: Maybe `Color.valueOf(0xFFC58AF9)`.  I'm asking, not that I know our best conventions.  I avoid this by always originating from a MaterialTheme.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
yschimke: Yep.  But it also has a much tighter window (x seconds) to complete, so can be problematic then.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileRenderer.kt
ataulm: I was about to add them but it's checked already in MessagingTileService. I added a comment to make it clearer.

Issue #219, opened 2022-02-21
AlwaysOnKotlin tests fail even with older checkouts


yschimke:
```
java.lang.IllegalStateException: No compose views found in the app. Is your Activity resumed?
	at androidx.compose.ui.test.TestContext.getAllSemanticsNodes$ui_test_release(TestOwner.kt:96)
	at androidx.compose.ui.test.SemanticsNodeInteraction.fetchSemanticsNodes$ui_test_release(SemanticsNodeInteraction.kt:79)
	at androidx.compose.ui.test.SemanticsNodeInteraction.fetchOneOrDie(SemanticsNodeInteraction.kt:145)
	at androidx.compose.ui.test.SemanticsNodeInteraction.fetchSemanticsNode(SemanticsNodeInteraction.kt:96)
```

Screen shows

![image](https://user-images.githubusercontent.com/231923/155004422-67fa0b38-cd85-4cf3-b885-db3ed7c3ff66.png)

This is both at head but also back at https://github.com/android/wear-os-samples/commit/15fdaf1a05300f9e1335b1451f81d90ac9a4a5d0

cc @alexvanyo @shumelchyk

yschimke:
Working fine for me locally on API 28.

yschimke:
Failing differently on API 26

```
com.example.android.wearable.wear.alwayson.AlwaysOnAppTests > textIsCorrectAfterReturningToActiveMode[test(AVD) - 8.0.0] FAILED 
	java.lang.AssertionError: Failed to assert the following: (Text + EditableText = [Draw count: 9])
	Semantics of the node:
Tests on test(AVD) - 8.0.0 failed: There was 1 failure(s).
Feb 21, 2022 4:56:34 PM com.google.testing.platform.RunnerImpl$Companion summarizeAndLogTestResult
SEVERE: Execute com.example.android.wearable.wear.alwayson.AlwaysOnAppTests.textIsCorrectAfterReturningToActiveMode: FAILED
java.lang.AssertionError: java.lang.AssertionError: Failed to assert the following: (Text + EditableText = [Draw count: 9])
Semantics of the node:
Node #35 at (l=75.0, t=170.0, r=205.0, b=199.0)px, Tag: 'drawCount'
Text = '[Draw count: 8]'
Actions = [GetTextLayoutResult]
Has 4 siblings
Selector used: (TestTag = 'drawCount')

at androidx.compose.ui.test.AssertionsKt.assert(Assertions.kt:270)
[...]

PR #501, opened 2022-08-09
Testing CI Build only

yschimke:
Simple PR branching off a previously successfully building commit https://github.com/android/wear-os-samples/commit/1c437722cb493890d76e7e54c1d2810113d9ac4d

If it fails, then likely the emulators or CI environment has changed.


yschimke:
cc @ataulm we have a flaky CI build, previous cause was not a code change, but new emulator images.

See https://github.com/android/wear-os-samples/issues/219#issuecomment-1047885080 for history.

PR #504, opened 2022-08-10
Use Compose colors for tiles sample

ataulm:
Instead of defining Tile colors explicitly, transform from a compose color palette, since that'll be the usual case (compose-based Wear app + tile + ...)

No changes outside this file (Colors is still accessed via `MessagingTileTheme.colors`).

### DIFF
--- a/WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileTheme.kt
+++ b/WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileTheme.kt
@@ -15,16 +15,35 @@
  */
 package com.example.wear.tiles.messaging
 
-import android.graphics.Color
-import androidx.wear.tiles.material.Colors
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.toArgb
+import androidx.wear.compose.material.Colors
 
 object MessagingTileTheme {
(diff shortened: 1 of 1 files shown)


yschimke (approved):
Nice

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileTheme.kt
```diff
@@ -15,16 +15,35 @@
...
+    surface = ColorPalette.darkBlue,
+    onSurface = ColorPalette.purple,
+)
+
+private object ColorPalette {
+    val purple = Color(0xFFC58AF9)
+    val darkBlue = Color(0xFF202124)
+}
+
+private fun Colors.toTileColors() = androidx.wear.tiles.material.Colors(
```
ataulm, line 44: seems like a good one for horologist-tiles?

yschimke: Yep - can move from here https://github.com/google/horologist/blob/6d119aae93d2b117d23cb39bfd9134f5b61d6f91/media-ui/src/main/java/com/google/android/horologist/media/ui/tiles/ToTileColors.kt

ataulm:
✌️

PR #513, opened 2022-08-11
Cleanup after spotless update

yschimke:
With the compose update we started getting this https://github.com/android/wear-os-samples/pull/507

```
> Could not create task ':app:spotlessKotlinCheck'.
   > Could not create task ':app:spotlessKotlin'.
      > KtLint versions < 0.46.0 not supported!
```

And pulling that thread we end up having to accept a bunch of new rules for ktlint, and spotless.  This is an example fixing up a PR using `./gradlew spotlessApply` and fixing up issues.


AlwaysOnKotlin/build.gradle
```diff
@@ -34,7 +34,7 @@ subprojects {
             targetExclude("$buildDir/**/*.kt")
             targetExclude("bin/**/*.kt")
 
-            ktlint(libs.versions.ktlint.get()).userData(["max_line_length" : "100"])
+            ktlint(libs.versions.ktlint.get())
```
ataulm, line 37: Was this the syntax to override it?

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/ControlDashboard.kt
```diff
@@ -63,7 +63,7 @@ fun ControlDashboard(
         circle = circle,
         mic = mic,
         play = play,
-        music = music,
+        music = music
```
ataulm, line 66: allow trailing comma was set to true for this sample, so how come it stripped it?

ataulm (approved):

AlwaysOnKotlin/build.gradle
yschimke: Yep, not now results in https://github.com/android/wear-os-samples/runs/7776351662?check_suite_focus=true

> UserData should not contain '.editorconfig' properties [max_line_length]. Such properties should be passed via the 'ExperimentalParams.editorConfigOverride' field. Note that this is only required for properties that (potentially) contain a value that differs from the actual value in the '.editorconfig' file.

WearSpeakerSample/wear/src/main/java/com/example/android/wearable/speaker/ControlDashboard.kt
yschimke: I can flip to false in .editorconfig so it's consistent. I'm about to go on holidays, so would like build green and passing and then forget about this :)

Or we can raise a bug, similar reports https://github.com/diffplug/spotless/issues/1193 and it's defined here and one more place https://github.com/pinterest/ktlint/blob/d64733ba1711c592fb78fe04d3472c5294004450/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/TrailingCommaOnCallSiteRule.kt#L290

yschimke:
Landing since, the build is broken anyway.

Issue #523, opened 2022-08-30
WatchFace Kotlin deprecated methods


lancelinksp:
Can someone please update the WatchFace Kotlin sample to use non deprecated methods
The sample includes several deprecated methods.

ComplicationUtils.kt includes deprecated methods:

1. constructor DefaultComplicationDataSourcePolicy(Int) is deprecated
2. setDefaultDataSourceType(ComplicationType): ComplicationSlotBuilder is deprecated

AnalogWatchCanvasRenderer.kt includes deprecated method:

1. CanvasRenderer is deprecated

lancelinksp:
Wow, that was quick! Thank you

PR #525, opened 2022-08-31
Update deprecate methods in Watchface sample

yschimke:
#Fixes https://github.com/android/wear-os-samples/issues/523


garanj (approved):
Nice!

Issue #435, opened 2022-06-29
RuntimePermissionsWear not able to start activities on the phone when compiling with SDK 33


Stjin:
When setting 

> compileSdk = 33

 and

> targetSdk = 33

in build.gradle the application won't launch the activity on the phone anymore after pressing the "phone information" button on the watch.

Probably due to the restrictions introduced in Android T.

Logs shown on phone:
```
W/PackageManager: Intent does not match component's intent filter: Intent { act=com.google.android.gms.wearable.BIND_LISTENER cmp=com.example.android.wearable.runtimepermissions/.IncomingRequestPhoneService }
W/PackageManager: Access blocked: ComponentInfo{com.example.android.wearable.runtimepermissions/com.example.android.wearable.runtimepermissions.IncomingRequestPhoneService}
W/ActivityManager: Unable to start service Intent { act=com.google.android.gms.wearable.BIND_LISTENER cmp=com.example.android.wearable.runtimepermissions/.IncomingRequestPhoneService } U=0: not found
```

As far as I see the intent filters are correct, I can't figure out why this is not working on Sdk 33.

shumelchyk:
Hi @Stjin, thank for reporting this issue!

The root cause of the issue is related to the behaviour changes to intent filters (https://developer.android.com/about/versions/13/behavior-changes-13#intent-filters) introduced in Android 13 (T).

There is known issue with `WearableListenerService` when `targetSDK` is set to `33` which is currently being looked into internally: https://issuetracker.google.com/235538840

This fix is required to make sample work with `targetSDK=33`, we'll keep the bug in the issue tracker updated.

Stjin:
Thanks for the update @shumelchyk !

shumelchyk:
@Stjin The fix for this issue has landed! 

The fix has been rolled out via GMSCore and will also require using the latest `com.google.android.gms:play-services-wearable:18.0.0` release. Changes to the `RuntimePermissionsWear` sample made in https://github.com/android/wear-os-samples/pull/554 - let me know if you still face any issues!

Note that you don’t need to add `BIND_LISTENER` manually, it has been deprecated for a long time and it continue to remain so (read more at https://android-developers.googleblog.com/2016/04/deprecation-of-bindlistener.html).

PR #554, opened 2022-09-20
Fix for Android13 compatibility issue (#435)

shumelchyk:
- Bump targetSDK to 33
- Use newer version of play-services-wearable library with added support for apps targeting Android13


pranavb (approved):
👍🏽

PR #573, opened 2022-10-19
Add a basic splashscreen

yschimke:
Add a splashscreen.

Also includes the code for Androidx Splashscreen, but not currently working.

https://user-images.githubusercontent.com/231923/196630435-e43e73fa-f52a-4957-9b23-eebdb58ae38b.mp4


yschimke:
Long term I believe androidx.splashscreen should be the right API to use https://developer.android.com/develop/ui/views/launch/splash-screen/migrate

But at the moment, it didn't just work for me, but I didn't figure out why.  Since I know some apps are using, I'm assuming I got it wrong, need a repro to raise a bug with.

shumelchyk:
> Long term I believe androidx.splashscreen should be the right API to use https://developer.android.com/develop/ui/views/launch/splash-screen/migrate

Yes, that's correct, especially taking into account any future platform releases that will be based on Android12+.

@yschimke  I managed to make splash screen working using compat library (see video), there are few pieces missing - I can commit as follow-up, or just share here. Let me know what works for you.

I came across the issue that androidx.splashscreen doesn't allow to specify smaller dimensions for the icon (as per https://developer.android.com/reference/kotlin/androidx/core/splashscreen/SplashScreen#specs), so I workaround it with using `@drawable/splash_background` instead of icon. This may lead to another subtle issues in future. I'll raise a FR for this.

[device-2022-10-31-202243.webm](https://user-images.githubusercontent.com/1803903/199103188-bda64e7d-1690-41b9-b035-5e08796b7617.webm)

yschimke:
Feel free to take over this PR and submit yourself.

Or if busy give me pointers.

shumelchyk (approved):

shumelchyk:
I pushed commit to your branch - ptal :)

yschimke:
Thanks, LGTM.

shumelchyk:
For the record, here is a feature request I raised to allow Wear OS compatible icon sizes in `androidx.splashscreen` library: https://issuetracker.google.com/256678037

yschimke:
Should be able to test with https://android-review.googlesource.com/c/platform/frameworks/support/+/2450245 soon

Issue #585, opened 2022-11-02
Remove wearApp configuration


luizgrp:
Is the `wearApp` configurations still relevant / needed in the following samples?

- [WearVerifyRemoteApp](https://github.com/android/wear-os-samples/blob/393bfc76d594c6b58f7a27f35a5e6145a23f276b/WearVerifyRemoteApp/Application/build.gradle#L57) 
- [RuntimePermissionsWear](https://github.com/android/wear-os-samples/blob/e21315e68784b20c3279561e1bf54b3753045967/RuntimePermissionsWear/Application/build.gradle#L64)
- [DataLayer](https://github.com/android/wear-os-samples/blob/393bfc76d594c6b58f7a27f35a5e6145a23f276b/DataLayer/Application/build.gradle#L75)

PR #586, opened 2022-11-02
#585 Remove dependency with wearApp configuration

luizgrp:
#### WHAT

Remove from the mobile app module, the dependency on the wear app module with `wearApp` configuration.

#### WHY

It does seem to be needed. Tested the app without it and it works fine.

The only place in the [documentation](https://developer.android.com/studio/build/dependencies#wear_apps) that I could find reference to this configuration seems to not explain why it is needed.

Related: #585

### DIFF
--- a/DataLayer/Application/build.gradle
+++ b/DataLayer/Application/build.gradle
@@ -71,6 +71,4 @@ dependencies {
     implementation libs.androidx.lifecycle.viewmodel.compose
     implementation libs.androidx.lifecycle.runtime.ktx
     implementation libs.playservices.wearable
-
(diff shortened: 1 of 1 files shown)


yschimke:
Aren't there 3 https://github.com/search?q=wearApp++repo%3Aandroid%2Fwear-os-samples+extension%3Agradle&type=Code&ref=advsearch&l=&l=

yschimke (approved):

luizgrp:
> Aren't there 3

I will remove the others in a separate PR, once I get to test the other samples.

PR #623, opened 2022-12-06
Use animated toggle controls in ToggleChips

shumelchyk:
In v1.1 `ToggleChip` support usage of animated toggle controls (`Checkbox`, `Switch` and `RadioButton`) that can be used instead of the static icons provided by `ToggleChipDefaults`.

### DIFF
--- a/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/benchmark/ScrollActivity.kt
+++ b/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/benchmark/ScrollActivity.kt
@@ -98,7 +98,7 @@ fun SomeScreenContent() {
                         checked = itemChecked,
                         modifier = Modifier.semantics {
                             this.contentDescription =
-                                "Item $item ${if (itemChecked) "Checked" else "Unchecked"}"
+                                if (itemChecked) "Checked" else "Unchecked"
                         }
                     )
                 }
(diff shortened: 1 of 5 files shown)


ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/benchmark/ScrollActivity.kt
```diff
@@ -95,9 +94,12 @@ fun SomeScreenContent() {
...
                 },
                 toggleControl = {
-                    Icon(
-                        imageVector = ToggleChipDefaults.switchIcon(checked = itemChecked),
-                        contentDescription = "Item $item ${if (itemChecked) "Checked" else "Unchecked"}"
+                    Switch(
+                        checked = itemChecked,
+                        modifier = Modifier.semantics {
+                            this.contentDescription =
+                                "Item $item ${if (itemChecked) "Checked" else "Unchecked"}"
```
yschimke, line 101: From the samples, it looks like the suggestion is to just use "Checked" and the context of the label still applies.

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/compose/compose-material/samples/src/main/java/androidx/wear/compose/material/samples/ToggleChipSample.kt?q=ToggleChip

Should we check with which it is? or just remove "Item $item " here?

shumelchyk: I'm leaning for latter, to remove "Item $item " - going to update now

PR #624, opened 2022-12-06
Demos for Outlined style for Chips and modifying Chip shape

shumelchyk:
Demos for a new features available in Compose for Wear OS version 1.1:
- Using Outlined style for Chips at `Dialogs` screen
- Example of modifying Chip shape at `Progress Indicators` screen

chip_styles_shape_demo


shumelchyk:
> Looking towards 2023, we should find a more cohesive solution for showcasing the components.
Yeah, definitely. There are lots of great examples in androidx itself, but Github samples are usually easier for exploration.

> The usage of outline chip here can't really give guidance on the correct use.
Do you refer to developer or design guidance (or both)?

yschimke:
> > Looking towards 2023, we should find a more cohesive solution for showcasing the components.
> > Yeah, definitely. There are lots of great examples in androidx itself, but Github samples are usually easier for exploration.
> 
> > The usage of outline chip here can't really give guidance on the correct use.
> 
> Do you refer to developer or design guidance (or both)?

Specifically Design guidance. Discussing offline.

Issue #584, opened 2022-11-02
Update to Compose October '22


JolandaVerhoef:
Would be great to update the Compose samples to use the Compose BOM. See https://github.com/android/user-interface-samples/pull/399 for an example of how to do such a migration.

shumelchyk:
Hi @JolandaVerhoef there is a planned update of the samples for Compose for Wear 1.1 version, so can be updated to BOM within these changes.

Can you clarify if BOM can be applied to Compose for Wear too?

cc @yschimke @BowerSteve since we briefly discussed it before.

JolandaVerhoef:
It's not included because the BOM only contains the libraries from the `androidx.compose` library group.

oas004:
According to [this thread](https://kotlinlang.slack.com/archives/C02GBABJUAF/p1668374681824269) in the Kotlin Slack, I understand it like there is no plan to make a BoM for Wear artefacts. Is that still something that should be waited for? Or is this something that can be contributed with the Compose BoM? :)

yschimke:
I think it's worthwhile to make the change, just not complete as there is no planned Wear BOM.

https://github.com/android/wear-os-samples/blob/main/ComposeAdvanced/gradle/libs.versions.toml

<img width="899" alt="image" src="https://user-images.githubusercontent.com/231923/208622838-cb0d3f3c-924b-4fca-ade2-6f8f52dc53e3.png">

PR #633, opened 2022-12-28
[ComposeStarter] Use December BoM for Compose dependencies

oas004:
Start using the December BoM for the Compose dependencies in the Compose Starter project.

Following up #631 relating to #584

### DIFF
--- a/ComposeStarter/app/build.gradle
+++ b/ComposeStarter/app/build.gradle
@@ -59,7 +59,11 @@ android {
 }
 
 dependencies {
+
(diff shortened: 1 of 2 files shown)


ComposeStarter/app/build.gradle
```diff
@@ -80,8 +84,13 @@ dependencies {
 
     // Testing
     testImplementation libs.junit
+    testImplementation composeBom
```
oas004, line 87: Maybe I don't need this actually since there is no compose test dependency for jvm tests 🤔

PR #634, opened 2022-12-28
[ComposeAdvanced] Start using the December BoM for Compose dependencies

oas004:
Start using the December BoM for Compose dependencies in the ComposeAdvanced sample.

Following up #631 relating to #584

### DIFF
--- a/ComposeAdvanced/app/build.gradle
+++ b/ComposeAdvanced/app/build.gradle
@@ -85,7 +85,10 @@ android {
 }
 
 dependencies {
+    def composeBom = platform(libs.androidx.compose.bom)
+
     // General compose dependencies
+    implementation composeBom
     implementation libs.androidx.activity.compose
     implementation libs.compose.ui.tooling.preview
     implementation libs.compose.foundation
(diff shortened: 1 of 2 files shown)


ComposeAdvanced/app/build.gradle
```diff
@@ -135,9 +138,11 @@ dependencies {
 
     // Testing
     testImplementation libs.junit
+    androidTestImplementation composeBom
     androidTestImplementation libs.test.ext.junit
     androidTestImplementation libs.test.espresso.core
     androidTestImplementation libs.compose.ui.test.junit4
+    debugImplementation composeBom
```
yschimke, line 145: Are you adding these because of failures?

oas004: No I was adding it since the compose.ui.tooling was using the BoM, so I thought I needed it since they basically does the same with `androidTestImplementation` [here](https://developer.android.com/jetpack/compose/setup#bom-version-mapping). I'm not 100% sure if it is needed though. I can test it out without it :)

oas004: When I am thinking about it, it does not quite make sense to add. Removed it.

oas004: Yeah I think you are correct :)

PR #671, opened 2023-01-26
[AlwaysOnKotlin] add missing dependency for release

CoreFloDev:
If `com.google.android.wearable:wearable:2.9.0` isn't specified then the ambient mode won't be triggered in release mode

### DIFF
--- a/AlwaysOnKotlin/compose/build.gradle
+++ b/AlwaysOnKotlin/compose/build.gradle
@@ -86,6 +86,9 @@ dependencies {
     implementation libs.wear.compose.material
     implementation libs.androidx.window
     implementation libs.compose.ui.test.manifest
+    
+    // Release mode doesn't work without that dependency
+    compileOnly "com.google.android.wearable:wearable:2.9.0"
 
     androidTestImplementation composeBom
     androidTestImplementation libs.kotlinx.coroutines.test


yschimke:
@shumelchyk Seems like an androidx bug if depending on androidx.wear:wear:1.2.0 is not enough to make this work.

AlwaysOnKotlin/compose/build.gradle
```diff
@@ -86,6 +86,9 @@ dependencies {
     implementation libs.wear.compose.material
     implementation libs.androidx.window
     implementation libs.compose.ui.test.manifest
+    
+    // Release mode doesn't work without that dependency
+    compileOnly "com.google.android.wearable:wearable:2.9.0"
```
yschimke, line 91: Should we make this implementation?

yschimke:
@CoreFloDev please sign the CLA.

yschimke:
Probably relates to this https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/wear/src/main/java/androidx/wear/ambient/WearableControllerProvider.java;l=87?q=com.google.android.wearable&ss=androidx%2Fplatform%2Fframeworks%2Fsupport:wear%2Fwear%2F

            throw new IllegalStateException(
                    "Could not find a required method for "
                            + "ambient support, likely due to proguard optimization. Please add "
                            + "com.google.android.wearable:wearable jar to the list of library jars"
                            + " for your project");

yschimke:
Also found this, so I wonder if it's the uses-library that is missing?

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/wear/src/main/java/androidx/wear/ambient/SharedLibraryVersion.java?q=com.google.android.wearable&start=51

        if (!PresenceHolder.PRESENT) {
            throw new IllegalStateException("Could not find wearable shared library classes. "
                    + "Please add <uses-library android:name=\"com.google.android.wearable\" "
                    + "android:required=\"false\" /> to the application manifest");

shumelchyk:
@yschimke I think it's likely the root cause.
@CoreFloDev can you please check if you see `Could not find a required method for ambient support` in the bugreport/logs?

> Probably relates to this https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:wear/wear/src/main/java/androidx/wear/ambient/WearableControllerProvider.java;l=87?q=com.google.android.wearable&ss=androidx%2Fplatform%2Fframeworks%2Fsupport:wear%2Fwear%2F
> 
> ```
>         throw new IllegalStateException(
>                 "Could not find a required method for "
>                         + "ambient support, likely due to proguard optimization. Please add "
>                         + "com.google.android.wearable:wearable jar to the list of library jars"
>                         + " for your project");
> ```

AlwaysOnKotlin/compose/build.gradle
CoreFloDev: that's not possible the implementation is only on the device from what I understood

CoreFloDev:
@shumelchyk nope nothing is visible on the logs

@yschimke Adding <uses-library android:name="com.google.android.wearable"
            android:required="false" /> doesn't solve the issue

It is quite funny that the messages that are supposed to help you aren't displayed, they probably get proguarded.

It is really only when I add the compileOnly dep that it works ok

yschimke:
@shumelchyk this being required seems like a bug.

https://developer.android.com/wear/releases#2022-february-16_wearable_support_library_v290

> Version 2.9.0 of the Wearable Support Library deprecates all remaining classes. Use the [Wear OS Jetpack libraries](https://developer.android.com/jetpack/androidx/releases/wear) instead.

CoreFloDev:
Why? `"com.google.android.wearable:wearable:2.9.0"` isn't the support library, it is a more low level artifact
The support library is `"com.google.android.support:wearable:2.9.0"`. I agree the naming is quite confusing but I guess it was making sense when only the support one was supposed to be used

PR #700, opened 2023-02-24
Testing splashscreen fixes

yschimke:
Not to land before [1.1.0-alpha01](https://developer.android.com/jetpack/androidx/releases/core) is published


yschimke:
@shumelchyk I can revert if needed, but I think it's the current approach.

PR #786, opened 2023-04-28
Update splash screen implementation

shumelchyk:
That's follow-up on https://github.com/android/wear-os-samples/pull/573, version `1.0.1` is the version that includes fix for https://issuetracker.google.com/256678037.

So now [core-splashscreen 1.0.1](https://developer.android.com/jetpack/androidx/releases/core#core-splashscreen-1.0.1) adds support for default Wear OS dimensions. With that, implementing splash screen can be easier, will give consistency across all Wear OS versions and meets design specification.


shumelchyk:
[splash_screen_demo.webm](https://user-images.githubusercontent.com/1803903/235209361-46ac133a-a8a3-4ebf-8c1f-418e581c8aa2.webm)

ComposeAdvanced/app/src/main/AndroidManifest.xml
```diff
@@ -11,21 +10,20 @@
...
 
     <application
+        android:name=".presentation.BaseApplication"
         android:allowBackup="true"
         android:fullBackupOnly="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
-        android:name=".presentation.BaseApplication"
-        android:supportsRtl="true"
-        android:theme="@style/SplashTheme">
```
yschimke, line 28: Does this remove it from maps?

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/benchmark/ScrollActivity.kt
```diff
@@ -47,13 +47,11 @@ class ScrollActivity : ComponentActivity() {
     internal lateinit var navController: NavHostController
 
     override fun onCreate(savedInstanceState: Bundle?) {
-        installSplashScreen()
```
yschimke, line 50: https://developer.android.com/develop/ui/views/launch/splash-screen/migrate

> Call installSplashScreen in the starting activity before calling super.onCreate().

yschimke (approved):
LGTM

ComposeAdvanced/app/src/main/AndroidManifest.xml
shumelchyk: I don't think we ever showed splash screen theme, I actually reverted this bit to keep splash theme under application's attribute, not activities'

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/benchmark/ScrollActivity.kt
shumelchyk: Good catch, fixed

PR #804, opened 2023-05-12
Add roundApi30 and roundApi33 tests to AlwaysOnKotlin CI workflow

alpergundogdu:


AlwaysOnKotlin/compose/build.gradle
```diff
@@ -93,7 +93,12 @@ android {
         device = "Wear OS Large Round"
         apiLevel = 30
         systemImageSource = "android-wear"
-      }
+      },
```
yschimke, line 96: comma not needed

yschimke: These are all method invocations.

yschimke:
Thanks for trying

```
   > System Image specified by roundApi33 does not exist.
     
     Try one of the following fixes:
     1. The image does not exist from android-wear for this architecture on the given apiLevel. However, other sources exist. Set systemImageSource to any of [google] to use.
     2. The system image does not presently exist for apiLevel 33. The latest available apiLevel is 30. Set apiLevel = 30 to use.
```

Let's leave the definition in for further testing, but just remove from the github workflow.

yschimke:
Just noticed the 30 also timed out again

alexvanyo:
If the `android-wear` image for API `33` is only available in the canary channel from `sdkmanager`, you can use a property `android.sdk.channel=3` in `gradle.properties` to allow fetching it.

yschimke:
Will land and tweak separately

PR #847, opened 2023-06-26
Use Horologist Compose Material

luizgrp:
### WHAT

Use Horologist Compose Material.

### WHY

Less verbose + a11y fixes.

### DIFF
--- a/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/landing/LandingScreen.kt
+++ b/ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/landing/LandingScreen.kt
@@ -30,7 +30,6 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextAlign
-import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import androidx.wear.compose.foundation.AnchorType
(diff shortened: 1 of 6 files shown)


ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/landing/LandingScreen.kt
```diff
@@ -87,28 +85,14 @@ fun LandingScreen(
...
                     Chip(
                         onClick = listItem.clickHander,
-                        label = {
-                            Text(
-                                listItem.name,
-                                maxLines = 1,
-                                overflow = TextOverflow.Ellipsis
-                            )
-                        },
-                        modifier = Modifier.fillMaxWidth()
```
yschimke, line 111: Isn't fill required?

yschimke (approved):
LGTM except for removing the fillMaxWidth

ComposeAdvanced/app/src/main/java/com/example/android/wearable/composeadvanced/presentation/ui/landing/LandingScreen.kt
luizgrp: It is [set](https://github.com/google/horologist/blob/564353a0232038764768a9d0e1874ea5065d8793/compose-material/src/main/java/com/google/android/horologist/compose/material/Chip.kt#L219) in Chip, since the beginning, but thinking in making it more flexible, maybe adding a flag to choose it to not be set. Thoughts?

yschimke: I'm not sure I've ever used without full width. So that's ok. It's probably more surprising it breaks the typical layout responsibilities.

Let's leave for now?

shumelchyk (approved):
LGTM, thanks Gustavo!

PR #954, opened 2023-11-05
Fix Tile previews

yschimke:
image


yschimke:
@garanj will land to fix an issue

PR #961, opened 2023-11-24
Adds example of complications for Watch Face Format

garanj:
Adds a basic example of using Complication data in a watch face


WatchFaceFormat/Complications/res/raw/watchface.xml
```diff
@@ -0,0 +1,722 @@
...
+      is shown also, and if not, then Title is shown, if present.
+    -->
+    <ComplicationSlot
+        x="56" y="168"
+        width="112" height="112"
+        slotId="1"
+        name="left_complication" displayName="Left"
+        supportedTypes="SHORT_TEXT">
+      <DefaultProviderPolicy
+          primaryProvider="com.google.android.gm/com.google.android.apps.gmail.wear.complications.UnreadEmailsComplicationService"
```
yschimke, line 372: Are these considered public API now?

If so, should we document the available ones?

yschimke (approved):

yschimke:
Looks great!

WatchFaceFormat/Complications/res/raw/watchface.xml
garanj: Do you mean for the `primaryProvider` or for `defaultSystemProvider`?

For `primaryProvider` (and secondary) you can add any component you like, but there is no guarantee it will be present.

I can add a link to the system providers: https://developer.android.com/reference/android/support/wearable/complications/SystemProviders.html#ANDROID_PAY ?

garanj: Added further documentation, in the first use of `DefaultProviderPolicy`:

https://github.com/android/wear-os-samples/pull/961/commits/833d83cc50354c92c9e1d3abf2b1502edad3e805

yschimke: I meant the Gmail activity

garanj: Ah ok. This was just meant to show an example of a complication component name, which deliberately may or may not be available, allowing demonstration of the `defaultSystemProvider`.

I don't think there's any documentation of these, I just picked out some available ones.

amoledwatchfaces (approved):

PR #965, opened 2023-12-06
Updates use of Template to use Parameter to ensure valid WFF

garanj:
`Template` must have at least 1 `Parameter` element, even though the renderer may work without.

Resources should be specified without the `@string` prefix.

### DIFF
--- a/WatchFaceFormat/SimpleDigital/res/raw/watchface.xml
+++ b/WatchFaceFormat/SimpleDigital/res/raw/watchface.xml
@@ -49,7 +49,7 @@
         <Text align="CENTER">
           <!-- Demonstrates using Android string resources for localization -->
           <Font family="SYNC_TO_DEVICE" size="36" color="#ffffffff">
-            <Template>@string/greeting</Template>
+            <Template>%s<Parameter expression="greeting"></Parameter></Template>
           </Font>
         </Text>
       </PartText>


WatchFaceFormat/SimpleDigital/res/raw/watchface.xml
```diff
@@ -49,7 +49,7 @@
         <Text align="CENTER">
           <!-- Demonstrates using Android string resources for localization -->
           <Font family="SYNC_TO_DEVICE" size="36" color="#ffffffff">
-            <Template>@string/greeting</Template>
+            <Template>%s<Parameter expression="greeting"></Parameter></Template>
```
yschimke, line 52: Not really about this PR, but I can't follow that this is correct from

https://developer.android.com/training/wearables/wff/group/part/text/formatter/template
https://developer.android.com/training/wearables/wff/common/attributes/arithmetic-expression

yschimke (approved):

WatchFaceFormat/SimpleDigital/res/raw/watchface.xml
garanj: I'll verify the snippets on those pages work correctly or not, and update them to cover the above. Thanks Yuri!

yschimke: I guess It's more a question of why

`<Parameter expression="greeting"></Parameter>` evaluates to R.string.greeting

While https://developer.android.com/training/wearables/wff/group/part/text/formatter/template shows

```
<Template>%s %d
    <Parameter expression="Current heart rate: " />
    <Parameter expression="[HEART_RATE]" />
</Template>
```

PR #981, opened 2024-01-05
focus on scroll view when creating watch config activity

derry701:
focus on scroll view when creating watch config activity to easily support rotary input from crown


yschimke:
Thanks

PR #985, opened 2024-01-11
Update Tiles previews

yschimke:


garanj (approved):
Nice, TilePreviewData looks handy!

yschimke:
@garanj I guess we should try in different versions of Android Studio. Then land?

yschimke:
I don't have an easy way to test, but if not supported it just shouldn't show it. The code should be safe.

yschimke:
cc @castedmo

PR #994, opened 2024-01-22
Run on faster ubuntu runners

yschimke:
https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/


yschimke:
These now show legitimate failures we should fix

yschimke:
@shumelchyk Going to land as it demonstrates the test failures.

PR #1011, opened 2024-02-28
Adds font scaling screenshot tests

garanj:
Adds font scaling tests


ComposeStarter/app/src/test/java/presentation/WearDevice.kt
```diff
@@ -58,6 +59,20 @@ enum class WearDevice(
         modelName = "Generic Large Round",
         screenSizePx = 454,
         density = 2.0f
+    ),
+    SamsungGalaxyWatch6SmallFont(
+        id = "galaxy_watch_6_small_font",
+        modelName = "Samsung Galaxy Watch 6 Large",
+        screenSizePx = 480,
+        density = 2.125f,
```
yschimke, line 67: BTW I struggled with this one with Robolectric. I think dp results are correct, but I think final px sizes are off...

yschimke (approved):

ComposeStarter/app/src/test/java/presentation/WearDevice.kt
garanj: Specific to the GW6?

yschimke: Anything with a quirky density

You can see the issue here. https://github.com/google/horologist/blob/main/sample/src/test/snapshots/images/com.google.android.horologist.screensizes_DevicesTest_screenshot%5B2%5D_samsunggalaxywatch6large.png

The results is correct. But the file resolution is wrong.

garanj: Huh, weird.

PR #1016, opened 2024-03-04
Choose a more problematic layout

yschimke:
Demonstrate the value of responsive layout, but choosing a layout that could break.


yschimke:
@garanj building on this, so will land before you get a chance to review.

PR #1020, opened 2024-03-05
Updates to use Horologist for ResponsiveTimeText and rememberResponsiveColumnState

garanj:
Updates to use Horologist for `ResponsiveTimeText` and `rememberResponsiveColumnState`


ComposeStarter/app/src/main/java/com/example/android/wearable/composestarter/presentation/MainActivity.kt
```diff
@@ -127,7 +128,16 @@ fun GreetingScreen(greetingName: String, onShowList: () -> Unit) {
...
 @Composable
 fun ListScreen() {
-    val columnState = rememberColumnState()
+    /*
+     * Specifying the types of items that appear at the start and end of the list ensures that the
+     * appropriate padding is used.
+     */
+    val columnState = rememberResponsiveColumnState(
+        contentPadding = ScalingLazyColumnDefaults.padding(
+            first = ScalingLazyColumnDefaults.ItemType.Text,
```
yschimke, line 137: I think we need to stop using `ListHeader { Text() }`

For follow up. As it pushes the text below where we want it.

yschimke (approved):

PR #1022, opened 2024-03-06
Add a sample dialog

yschimke:


ComposeStarter/app/src/main/java/com/example/android/wearable/composestarter/presentation/MainActivity.kt
```diff
@@ -187,6 +206,48 @@ fun Greeting(greetingName: String) {
...
+@Composable
+fun SampleDialog(
+    showDialog: Boolean,
+    onDismiss: () -> Unit,
+    onCancel: () -> Unit,
+    onOk: () -> Unit
+) {
+    val state = rememberResponsiveColumnState()
+
+    Dialog(
```
garanj, line 218: What's the reasoning behind using standard `Dialog` with Horologist `AlertContent`, instead of Horologist `AlertDialog`?

garanj:
> @garanj @kul3r4 I am not sure where the line is, is this getting too complicated? Or is it useful to show a Dialog sample.

I agree that it's useful, I don't think that it adds too much to the complexity. I do agree that we should draw a line in this starter sample, but IMO this seems useful.

garanj (approved):

ComposeStarter/app/src/main/java/com/example/android/wearable/composestarter/presentation/MainActivity.kt
yschimke: Screenshot tests. I think AlertDialog needs some tweaks somewhere to support screenshotting a dialog.  I'll follow up.

PR #1028, opened 2024-03-08
Adds splashscreen for Wear 3

kul3r4:
Add splashscreen for Wear 3 by following guidance from https://developer.android.com/reference/androidx/core/splashscreen/SplashScreen.
It adds a theme file and a splashscreen.xml so that we can define sizes (otherwise the icon is too big.)
Also adds adaptive icons both for launcher and splashscreen.


kul3r4:
@ithinkihaveacat take a look for the addition of the adaptive icons if it looks good.
@garanj FYI

ComposeStarter/app/src/main/res/values/themes.xml
```diff
@@ -0,0 +1,27 @@
...
+     Unless required by applicable law or agreed to in writing, software
+     distributed under the License is distributed on an "AS IS" BASIS,
+     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+     See the License for the specific language governing permissions and
+     limitations under the License.
+-->
+<resources xmlns:tools="http://schemas.android.com/tools">
+    <style name="Theme.App" parent="@android:style/Theme.DeviceDefault" />
+
+    <style name="Theme.App.Starting" parent="Theme.SplashScreen">
```
yschimke, line 20: Could this extend Theme.SplashScreen.IconBackground from https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core-splashscreen/src/main/res/values/styles.xml?q=Theme.SplashScreen.IconBackground 

And could we use the default values from https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core-splashscreen/src/main/res/values-watch/dimens.xml;l=26

yschimke (approved):
LGTM - correct based on docs for splashscreen, but I am wondering whether we can simplify that to use more defaults?

kul3r4:
The issue when not specifying 
```
 android:width="@dimen/splash_screen_icon_size"
        android:height="@dimen/splash_screen_icon_size"
```
even if you use `parent="Theme.SplashScreen.IconBackground"`, it's that you will get a different icon from the launcher in size. Attaching video.
[Screen_recording_20240310_210018.webm](https://github.com/android/wear-os-samples/assets/820891/2904a850-2575-4a78-9676-9f8559a480f0)

Was the suggestion to remove dimens.xml?

yschimke:
I think we should fix the library to use the correct defaults!

But for now, this is correct.

yschimke:
Going to land, to play around with.  Any further comments in a follow up?

Hope this is ok @kul3r4

yschimke:
@kul3r4 from https://issuetracker.google.com/issues/256678037 - it seems like the values 72/90 in https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:core/core-splashscreen/src/main/res/values-watch/dimens.xml;l=26  are deliberate.

Should we reopen this issue?

PR #1050, opened 2024-03-21
Add deprecation warning to deprecated samples

kul3r4:
Add deprecation warning to samples. Next step is to remove those samples from main.


ComposeAdvanced/README.md
```diff
@@ -1,3 +1,5 @@
+This project is now deprecated and unmaintained.
```
yschimke, line 1: Is it worth giving either 
a) a brief reasons
b) a link

If we remove from main, probably not.  But figured worth asking.

kul3r4: you are totally right, worth linking to Horologist and to ComposeStarter, will add as part of this PR.

kul3r4:
@yschimke added links in every Readme pointing to other samples.

PR #1067, opened 2024-04-17
Adds gradle configuration for SimpleDigital watch face

garanj:
Adds gradle configuration for `SimpleDigital` watch face example.

Once added, other watch faces to follow, plus updated documentation.


WatchFaceFormat/SimpleDigital/watchface/build.gradle.kts
```diff
@@ -0,0 +1,37 @@
...
+        applicationId = "com.example.simpledigital"
+        minSdk = 33
+        targetSdk = 33
+        versionCode = 1
+        versionName = "1.0.0"
+    }
+
+    buildTypes {
+        release {
+            isMinifyEnabled = true
```
yschimke, line 34: Do you want to add `isShrinkResources = false`. I'm assuming we should disable? Probably with a comment.

yschimke, line 34: Does this build ok without a signing key?

garanj: It does build, though it doesn't deploy. I've added a comment to show that the developer should add a signingConfig.

garanj: I've sent a question to the watch face team. Will update on their response.

yschimke: I assume the build system won't understand that WFF references resources.

PR #1078, opened 2024-04-25
Renames to Complications

garanj:
Changes the name of the project to Complications, for simplicity.


yschimke:
Is this an importable sample, will that be renamed?

garanj:
> Is this an importable sample, will that be renamed?

Imported into g3?

yschimke:
Android studio

garanj:
> Android studio

Got it! Thanks!!

kul3r4 (approved):

PR #1102, opened 2024-05-20
Release build of Watchfaces

yschimke:
Use a debug signing key so signed APKs are generated for local testing


garanj (approved):
Thanks Yuri!

Issue #1041, opened 2024-03-17
Refresh WearSpeakerSample


kul3r4:
Check if it can use Horologist.
Check logic for BT.

PR #1103, opened 2024-05-20
Simplify layout of Speaker Sample

kul3r4:
Further cleanup for #1041
This simplifies the layout removing ConstraintLayout.


yschimke:
Can we remove it from the gradle file?

  implementation libs.androidx.constraintlayout.compose

yschimke (approved):
nice simplification

PR #1105, opened 2024-05-23
Datalayer improvements

kul3r4:
Various improvements on the Wear app:

- Uses Horologist for AppScaffold and ScreenScaffold: this allows to remove handling of TimeText and PositionIndicator.
- Removes vignette
- Removes mentioning NodeClient which is not used, only CapabilityClient is used for discovery
- Adds adaptive icons (also on the phone app so that they match)
- Uses Chips instead of buttons
- Uses latest Preview lib


WearSpeakerSample/README.md
```diff
@@ -9,6 +9,11 @@ play the recorded voice, if the wearable device is connected to a speaker
 
 This sample is also written entirely with Jetpack Compose, using [Wear-specific components provided
 by AndroidX](https://developer.android.com/jetpack/androidx/releases/wear-compose).
+<<<<<<< HEAD
```
garanj, line 12: I think there are some merge remnants here?

garanj (approved):
Great! I added one small comment about one of the READMEs

WearSpeakerSample/README.md
kul3r4: Thanks!

DataLayer/Wearable/build.gradle
```diff
@@ -92,4 +93,15 @@ dependencies {
...
+
+  // Horologist for correct Compose layout
+  implementation libs.horologist.composables
+  implementation libs.horologist.compose.layout
+  implementation libs.horologist.compose.material
+
+  coreLibraryDesugaring libs.desugar.jdk.libs
+
+  // Preview Tooling
+  implementation libs.compose.ui.tooling.preview
```
yschimke, line 105: I think, but always get this wrong.  This should be the debugImplementation. And tooling without preview should be implementation.

yschimke (approved):

kul3r4:
Related to #1039

PR #1116, opened 2024-06-05
Update sample with more documentation and remove Toasts

kul3r4:
- remove Toasts and create 2 separate screens for showing nodes
- Update ReadMe with more details and point to Horologist for specific use cases


DataLayer/Application/src/main/java/com/example/android/wearable/datalayer/MainActivity.kt
```diff
@@ -85,32 +94,47 @@ class MainActivity : ComponentActivity() {
...
                 }
             }
         }
 
         setContent {
             MaterialTheme {
+                val coroutineScope = rememberCoroutineScope()
+                var apiAvailable by remember { mutableStateOf(false) }
+                LaunchedEffect(Unit) {
+                    coroutineScope.launch {
```
yschimke, line 129: You don't need a coroutineScope here since you have launched effect

DataLayer/Wearable/src/main/java/com/example/android/wearable/datalayer/ConnectedNodesScreen.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.google.android.horologist.compose.layout.rememberResponsiveColumnState
+
+@OptIn(ExperimentalHorologistApi::class)
+@Composable
+fun ConnectedNodesScreen(
+    viewModel: NodesViewModel
+) {
+    val state by viewModel.state.collectAsStateWithLifecycle()
+
+    val columnState = rememberResponsiveColumnState(
```
yschimke, line 34: I'd suggest moving this down into NodesScreen. It knows about the content structure, and doesn't need a param.

DataLayer/gradle/libs.versions.toml
```diff
@@ -2,7 +2,7 @@
 android-gradle-plugin = "8.4.1"
 androidx-activity = "1.9.0"
 androidx-compose-bom = "2024.05.00"
-androidx-lifecycle = "2.8.1"
```
yschimke, line 5: Worth a comment, otherwise we will likely revert.

Also I think maybe we should consider moving past this, by updating to beta since they are getting close to stable.

DataLayer/Wearable/src/main/java/com/example/android/wearable/datalayer/NodesViewModel.kt
```diff
@@ -0,0 +1,99 @@
...
+    /**
+     * Collects the capabilities for all nodes that are reachable using the [CapabilityClient].
+     *
+     * [CapabilityClient.getAllCapabilities] returns this information as a [Map] from capabilities
+     * to nodes, while this function inverts the map so we have a map of [Node]s to capabilities.
+     *
+     * This form is easier to work with when trying to operate upon all [Node]s.
+     */
+    private suspend fun getCapabilitiesForReachableNodes(): Map<Node, Set<String>> =
+        capabilityClient.getAllCapabilities(CapabilityClient.FILTER_REACHABLE)
```
yschimke, line 68: This call can fail, so we should probably catch the ApiExceptions and log?

DataLayer/Wearable/src/main/java/com/example/android/wearable/datalayer/NodesScreen.kt
```diff
@@ -0,0 +1,55 @@
...
+    nodes: Set<Node>,
+    modifier: Modifier = Modifier
+) {
+    ScreenScaffold(scrollState = columnState) {
+        ScalingLazyColumn(
+            columnState = columnState,
+            modifier = modifier
+        ) {
+            item {
+                ListHeader {
```
yschimke, line 42: ResponsiveListHeader, and contentPadding?

DataLayer/Wearable/src/main/java/com/example/android/wearable/datalayer/MainApp.kt
```diff
@@ -46,99 +49,124 @@ import com.google.android.horologist.compose.layout.ScreenScaffold
...
-            ) {
-                item {
-                    Chip(
-                        label = stringResource(id = R.string.query_other_devices),
-                        onClick = onQueryOtherDevicesClicked
-                    )
+        ScalingLazyColumn(
+            columnState = columnState,
+            modifier = Modifier
+                .fillMaxSize()
```
yschimke, line 102: fillMaxSize is not needed

DataLayer/Wearable/src/main/java/com/example/android/wearable/datalayer/MainApp.kt
```diff
@@ -46,99 +49,124 @@ import com.google.android.horologist.compose.layout.ScreenScaffold
...
+        }
+    }
+}
+
+@OptIn(ExperimentalHorologistApi::class)
+@Composable
+fun MainScreen(
+    onShowNodesList: () -> Unit,
+    onShowCameraNodesList: () -> Unit,
+    image: Bitmap?,
```
yschimke, line 84: Bitmap might not be a good type on a Composable function.  We should probably introduce some data class and mark it Stable?

yschimke (approved):
Nothing blocking

DataLayer/gradle/libs.versions.toml
kul3r4: I was getting a compatibility issue and this is why I moved to 2.7 but it's now solved, so I will revert this back

DataLayer/Wearable/src/main/java/com/example/android/wearable/datalayer/NodesViewModel.kt
kul3r4: Under which circumstances can fail?

yschimke: It's not likely, more about correct code.

https://developers.google.com/android/reference/com/google/android/gms/wearable/WearableStatusCodes

I think any call can technically fail with TIMEOUT, REMOTE_EXCEPTION, INTERNAL_ERROR.

#tioli

yschimke: We should consider these a remote call to another process on the device.

DataLayer/Wearable/src/main/java/com/example/android/wearable/datalayer/MainActivity.kt
```diff
@@ -41,77 +34,17 @@ class MainActivity : ComponentActivity() {
...
 
+        var nodesViewModel = NodesViewModel(capabilityClient)
+
         setContent {
             MainApp(
                 events = clientDataViewModel.events,
                 image = clientDataViewModel.image,
-                onQueryOtherDevicesClicked = ::onQueryOtherDevicesClicked,
-                onQueryMobileCameraClicked = ::onQueryMobileCameraClicked
+                nodesViewModel = nodesViewModel
```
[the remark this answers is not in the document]
kul3r4: Good point, at the moment the `clientDataViewModel` is extending the various listeners (message, capability etc.) and overrides the `onchanged` but we also have a `DataLayerListenerService` that implements the `WearableListenerService` and there again the `onchanged` methods are overriden. Perhaps we should try to keep only one place where those are overriden?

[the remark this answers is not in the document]
kul3r4: Moved the nodesViewModel creation, e.g.
```
fun CameraNodesScreen(
    viewModel: NodesViewModel = viewModel(factory = NodesViewModel.Factory)
```

kul3r4:
Related to #1039

DataLayer/Wearable/src/main/java/com/example/android/wearable/datalayer/MainActivity.kt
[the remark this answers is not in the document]
kul3r4: For the other part of the comment related to creation of the ViewModel, tracking it in #1119

PR #1131, opened 2024-06-18
Adds additional weather conditions

garanj:
Adds further weather conditions for WFF weather sample


WatchFaceFormat/Weather/watchface/src/main/res/raw/watchface.xml
```diff
@@ -74,15 +74,11 @@
                 </Group>
                 <Group height="47" name="CurrentConditionIcon" width="47" x="215" y="0">
                   <Condition>
-                    <!--
-                      When using CONDITION, it is a requirement that all
```
yschimke, line 78: What did this mean? And what was it referencing? Sounds scary.

WatchFaceFormat/Weather/watchface/src/main/res/raw/watchface.xml
```diff
@@ -74,15 +74,11 @@
...
-                      potential values are correctly handled. See:
-                      https://developer.android.com/docs/quality-guidelines/wear-app-quality
-                    -->
                     <Expressions>
+                      <Expression name="is_unknown">[WEATHER.CONDITION] == 0</Expression>
                       <Expression name="is_clear">[WEATHER.CONDITION] == 1</Expression>
                       <Expression name="is_cloudy">[WEATHER.CONDITION] == 2</Expression>
-                      <Expression name="is_fog">([WEATHER.CONDITION] == 3) || ([WEATHER.CONDITION] == 13)</Expression>
+                      <Expression name="is_mist">[WEATHER.CONDITION] == 13</Expression>
```
yschimke, line 81: reorder?

WatchFaceFormat/Weather/watchface/src/main/res/raw/watchface.xml
```diff
@@ -74,15 +74,11 @@
                 </Group>
                 <Group height="47" name="CurrentConditionIcon" width="47" x="215" y="0">
                   <Condition>
-                    <!--
-                      When using CONDITION, it is a requirement that all
-                      potential values are correctly handled. See:
-                      https://developer.android.com/docs/quality-guidelines/wear-app-quality
-                    -->
                     <Expressions>
+                      <Expression name="is_unknown">[WEATHER.CONDITION] == 0</Expression>
```
yschimke, line 78: What about > 22 also?

yschimke (approved):

WatchFaceFormat/Weather/watchface/src/main/res/raw/watchface.xml
yschimke: That would have been painful.  Why can't I have a WF that only shows rain, and is normal for other things.

garanj: You can 😀

WatchFaceFormat/Weather/watchface/src/main/res/raw/watchface.xml
lucianbc: It is not enforced, but you shouldn't get values > 22. If you get, it's a problem with the weather provider and in this particular example, nothing would be rendered. A developer might add a `Default` node at the same level with the `Compare` ones, but that is not needed. Should we add a comment about this?

garanj: I think there's the potential to overcomplicate the sample: The provider shouldn't provide > 22, I think the sample should keep it simple and assume it will do the right thing, which is generally reasonable?

WatchFaceFormat/Weather/watchface/src/main/res/raw/watchface.xml
garanj: @lucianbc is this something you can reorder on the source of truth?

WatchFaceFormat/Weather/watchface/src/main/res/raw/watchface.xml
yschimke: We are literally doing this right now.  So the values changing does happen.

I'd consider it maybe more like an exhaustive else.

But feel free to resolve.

garanj: Default case added

PR #1132, opened 2024-06-19
Use BitmapFonts for weather icons instead of multiple conditions with Images

amoledwatchfaces:
**We can use BitmapFonts element to specify set of weather icons for day / night.** 

This allows watchface.xml to be more readable & cleaner. It also helps reducing code by 50% and helps renderer as it does not need to compare many conditions. All while maintaining same functionality.


yschimke:
Nice, seems like a smart change. I'll leave for @garanj to approve.

garanj:
Hi @amoledwatchfaces apologies for the delay, you can see that the recent changes to the conditions enum have been reverted, so the highest value is now 15 - WINDY.

Would you be able to revisit your PR so that it only covers those cases, not up to 22?

WatchFaceFormat/Weather/watchface/src/main/res/raw/watchface.xml
```diff
@@ -75,114 +134,29 @@
...
-                    <Compare expression="is_tropical_cyclone">
-                      <PartImage height="47" width="47" x="0" y="0">
-                        <Image resource="ic_tropical_cyclone"/>
-                      </PartImage>
+                    <Compare expression="is_day">
+                      <PartText name="current_condition_icon" height="47" width="47" x="0" y="0">
+                        <Text>
+                          <BitmapFont family="weather_icons_day" size="47">
+                            <Template><![CDATA[%s]]>
+                              <Parameter expression="[WEATHER.CONDITION]"/>
```
garanj, line 144: `[WEATHER.CONDITION] &lt;= 15 : [WEATHER.CONDITION] : 0` in case the value is somehow out of the expected range?

amoledwatchfaces: Added in https://github.com/android/wear-os-samples/pull/1132/commits/059ced6fb668edbb8b8ab38e9cb6957b3275c08c

amoledwatchfaces:
not `[WEATHER.CONDITION] &lt;= 15 : [WEATHER.CONDITION] : 0`
but `[WEATHER.CONDITION] &lt;= 15 ? [WEATHER.CONDITION] : 0`

garanj (approved):

PR #1146, opened 2024-07-09
Use kotlin build.gradle in ComposeStarter

### LINKED ISSUE #1134: Migrate build scripts to Kotlin DSL
Most of the samples use Groovy build scripts, instead of Kotlin. Kotlin [is recommended](https://developer.android.com/build/migrate-to-kotlin-dsl) because it "is more readable and offers better compile-time checking and IDE support."

The affected samples are:

- [ ] Complications
- [x] ComposeStarter
- [x] DataLayer
- [ ] WatchFaceKotlin
- [ ] WearOAuth
- [ ] WearSpeakerSample
- [ ] WearTilesKotlin

kul3r4:
Migrate guild files from Groovy to Kotlin in ComposeStarter.
Partially fixes #1134


ComposeStarter/app/build.gradle.kts
```diff
@@ -0,0 +1,127 @@
...
+
+    val composeBom = platform(libs.androidx.compose.bom)
+
+    // General compose dependencies
+    implementation(composeBom)
+    implementation(libs.androidx.activity.compose)
+    implementation(libs.androidx.splashscreen)
+
+    // Compose for Wear OS Dependencies
+    // NOTE: DO NOT INCLUDE a dependency on androidx.compose.material:material.
```
yschimke, line 80: If we really believe this we should probably change this to a Lint?

Or fail the build if present. I can try the latter if you agree.

For follow up.

ComposeStarter/app/build.gradle.kts
```diff
@@ -0,0 +1,127 @@
...
+    implementation(libs.wear.compose.foundation)
+    implementation(libs.androidx.material.icons.core)
+
+    // Horologist for correct Compose layout
+    implementation(libs.horologist.composables)
+    implementation(libs.horologist.compose.layout)
+    implementation(libs.horologist.compose.material)
+
+    // Preview Tooling
+    implementation(libs.compose.ui.tooling.preview)
```
yschimke, line 97: For follow up - these are confusing because name is truncated and missing "wear".

```
androidx-compose-ui-tooling = { module = "androidx.wear.compose:compose-ui-tooling", version.ref = "compose-ui-tooling" }
```

yschimke (approved):

ComposeStarter/app/build.gradle.kts
kul3r4: Thanks that would be great!

ComposeStarter/app/build.gradle.kts
kul3r4: Happy to address in another PR and apply to all samples, what about something like `androidx-wear-compose-ui-tooling`?

kul3r4:
Thanks @yschimke for the review! I just added also https://github.com/android/wear-os-samples/pull/1146/files#diff-97720b8076bde3b9c20d2082ff624094b85cf52190de2198cde1d981198e75b4R16, can you PTAL?

yschimke (approved):

PR #1162, opened 2024-07-31
Migrate to Jetpack Tile previews

ithinkihaveacat:
This is a mostly mechanical change however there is one meaningful difference: previously the `ImageResource` passed to `onTileResourceRequest()` was passed through `bitmapToImageResource()`, where the `Bitmap` was extracted from the context via `BitmapFactory.decodeResource()`. I think this was mostly to demonstrate how to process a raw `Bitmap`.

However, with the new preview library, it's not possible to access the context (I think?), so we bypass this transformation step and instead generate the `ImageResource` via `drawableResToImageResource()` (which takes a resource id).

I think this is fine since the main point is to generate a preview rather than transform images from one format to another (the previews themselves are identical) but others may disagree!

Before ("compose"):

Screenshot 2024-07-31 at 15 09 10

Before ("Wear tile services"):

Screenshot 2024-07-31 at 15 09 19

After:

Screenshot 2024-07-31 at 15 10 45


yschimke:
The description doesn't match the PR

> However, with the new preview library, it's not possible to access the context

When it clearly is.

yschimke (approved):
Nice

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileLayout.kt
```diff
@@ -100,83 +96,77 @@ private fun searchLayout(
...
-private fun MessageTilePreview(context: Context): TilePreviewData {
+@Preview(device = WearDevices.SMALL_ROUND)
+@Preview(device = WearDevices.LARGE_ROUND)
+private fun messagingTilePreview(context: Context): TilePreviewData {
     val state = MessagingTileState(MessagingRepo.knownContacts)
-    return TilePreviewData(onTileResourceRequest = resources {
-        addIdToImageMapping(
-            state.contacts[1].imageResourceId(),
-            bitmapToImageResource(
-                BitmapFactory.decodeResource(context.resources, R.drawable.ali)
```
yschimke, line 111: Using `drawableResToImageResource` seems like an improvement.

You could consider inlining it if you want to minimise horologist usage here.

```
public fun drawableResToImageResource(
    @DrawableRes id: Int,
): ImageResource =
    ImageResource.Builder()
        .setAndroidResourceByResId(
            AndroidImageResourceByResId.Builder()
                .setResourceId(id)
                .build(),
        )
        .build()
```

garanj (approved):
Great!

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileLayout.kt
ithinkihaveacat: I was considering an extension function to enable `addIdToImageMapping()` to take a DrawableRes _or_ ImageResource:

```kotlin
fun ResourceBuilders.Resources.Builder.addIdToImageMapping(
    id: String,
    @DrawableRes resId: Int
): ResourceBuilders.Resources.Builder = addIdToImageMapping(
    id, ResourceBuilders.ImageResource.Builder()
        .setAndroidResourceByResId(
            ResourceBuilders.AndroidImageResourceByResId.Builder()
                .setResourceId(resId)
                .build()
        )
        .build()
)
```

Not sure whether this is idiomatic (versus the function approach) or whether it should be called `addIdToResourceMapping()`, etc. so was leaving it for a future PR.

ithinkihaveacat:
> The description doesn't match the PR
> 
> > However, with the new preview library, it's not possible to access the context
> 
> When it clearly is.

I don't follow this! This is the issue I was getting with a more mechanical transformation:

Screenshot 2024-07-31 at 13 36 11

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileLayout.kt
ithinkihaveacat: Added enhanced `addIdToImageMapping()` in https://github.com/android/wear-os-samples/pull/1164.

ithinkihaveacat:
> > The description doesn't match the PR
> > > However, with the new preview library, it's not possible to access the context
> > 
> > 
> > When it clearly is.
> 
> I don't follow this! This is the issue I was getting with a more mechanical transformation:
> 
> Screenshot 2024-07-31 at 13 36 11

Oh, there's a bug in the lint check. It's been reported.

PR #1164, opened 2024-08-02
Support drawables "directly"

ithinkihaveacat:
1. Is this the right API? (c.f. Horologist's [drawableResToImageResource()](https://github.com/google/horologist/blob/27802d54d07ca17061b6a13185c77586bc6c5f02/tiles/src/main/java/com/google/android/horologist/tiles/images/DrawableResToImageResource.kt#L28) approach.)
2. Is this the right place to put it? (Should it be in `tools/`?)
3. What happens if Tiles just adds this in a future release (as a regular method)? This code will error at compile time but … that's probably okay?

### DIFF
--- a/WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileLayout.kt
+++ b/WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileLayout.kt
@@ -16,8 +16,10 @@
 package com.example.wear.tiles.messaging
 
 import android.content.Context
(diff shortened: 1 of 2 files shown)


ithinkihaveacat:
Another option is to not overload `addIdToImageMapping()`, and call it `addIdToResourceMapping()` instead.

garanj (approved):

ithinkihaveacat:
Any thoughts on this @yschimke?

yschimke (approved):
I wouldn't overthink it.

PR #1173, opened 2024-08-11
extract common preview into custom annotation

hiroaki404:
I extracted common preview into custom annotation.


yschimke (approved):
Nice. I guess small risk it reduces slightly the supported IDEs, but seems like a good cleanup.

PR #1184, opened 2024-09-10
Update watchface.xml

amoledwatchfaces:
Removed **@drawable/** references so sample watch face can pass memory-footprint validation.


yschimke:
Is this a bug in the validator? Or it's a valid difference?

amoledwatchfaces:
> Is this a bug in the validator? Or it's a valid difference?

I think a valid difference now: 
https://developer.android.com/training/wearables/wff/group/part/image/image#required-attributes

> Note: You shouldn't include drawable/ or @drawable/ or the file suffix in the resource attribute value.

With the resource="@drawable/xxx", memory-footprint can fail like here: https://github.com/google/watchface/issues/48#issuecomment-2337324762

PR #1193, opened 2024-10-14
Adds use of validator plugin

garanj:
Adds WFF validation to the WFF samples:

1. By default, when running `gradlew assembleDebug` or `gradlew bundleDebug`, validation will be applied.
2. Separately, a `gradlew validateWff` task is available, for only validating the WFF XML
3. A further task `gradlew validateWffAndInstall` is available which will build, install and set the watch face for the project (with optional device specified using `--device <device-name>`.

Currently this is achieved through a plugin bundled with the samples, but this could be moved to a standalone published plugin.


WatchFaceFormat/Weather/watchface/build.gradle.kts
```diff
@@ -14,14 +14,18 @@
...
+    // Use the locally-defined validator to demonstrate validation on-build.
+    id("com.google.wff.validatorplugin")
 }
 
 android {
     namespace = "com.example.weather"
     compileSdk = 34
 
     defaultConfig {
+        applicationId = "com.example.weather"
```
yschimke, line 28: WAT?
WAT?

WatchFaceFormat/validator-plugin/settings.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+pluginManagement {
+    repositories {
+maven { url = uri("") }
```
yschimke, line 20: WAT?

WatchFaceFormat/validator-plugin/settings.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+        gradlePluginPortal()
+        google()
+        mavenCentral()
+    }
```
yschimke, line 24: What about adding

```
  versionCatalogs { create("libs") { from(files("../gradle/libs.versions.toml")) } }
```

WatchFaceFormat/validator-plugin/settings.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+dependencyResolutionManagement {
+    repositories {
+maven { url = uri("") }
```
yschimke, line 29: WAT?

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorExec.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import org.gradle.api.GradleException
+import org.gradle.api.Project
+import java.nio.file.Path
+
+
+/**
+ * Runs the validator against WFF XML files found in the project structure, matching any files:
+ *
+ *    res\raw*\*.xml
```
yschimke, line 25:
```suggestion
 *    res/raw*/*.xml
```

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorExec.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+internal fun validateWffFiles(project: Project, version: Int) {
+    val resourcePath = Path.of(project.projectDir.absolutePath, "src", "main", "res")
```
yschimke, line 28: I suspect this doesn't track the files as inputs correctly, but let me confirm.

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/WffValidatorPlugin.kt
```diff
@@ -0,0 +1,84 @@
...
+const val VALIDATE_TASK = "validateWff"
+const val DOWNLOAD_VALIDATOR_TASK = "downloadWffValidator"
+const val INSTALL_TASK = "validateWffAndInstall"
+
+class WffValidatorPlugin : Plugin<Project> {
+    private lateinit var manifestPath: String
+
+    override fun apply(project: Project) {
+        project.tasks.register(DOWNLOAD_VALIDATOR_TASK) { task ->
+            task.doLast {
```
yschimke, line 40: Something to consider

https://github.com/liutikas/gradle-best-practices?tab=readme-ov-file#create-custom-tasks

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorExec.kt
yschimke: Sort of this, https://tomgregory.com/gradle/gradle-task-inputs-and-outputs/

yschimke: But maybe a follow up PR?

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorDownload.kt
```diff
@@ -0,0 +1,70 @@
...
+
+private const val VALIDATOR_URL =
+    "https://github.com/google/watchface/releases/download/release/dwf-format-2-validator-1.0.jar"
+private const val VALIDATOR_FILE_NAME = "validator.jar"
+
+/**
+ * Downloads the WFF validator for use in the build process.
+ */
+internal fun downloadFileToPath(filePath: Path, url: String) {
+    val client = HttpClient()
```
yschimke, line 36: Should use expectSuccess here.

```
    val client = HttpClient {
        expectSuccess = true
    }
```

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorDownload.kt
```diff
@@ -0,0 +1,70 @@
...
+    val file = filePath.toFile()
+
+    runBlocking {
+        client.prepareGet(url).execute { httpResponse ->
```
yschimke, line 40: If you want to change to ktor 3, I think this works.

```
    runBlocking {
        client.get(url).bodyAsChannel().copyAndClose(filePath.toFile().writeChannel())
    }
```

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorDownload.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+}
+
+internal fun ensureValidatorDownloaded(project: Project) {
+    prepareValidatorDirectory(project)
+    val validatorJarPath = getValidatorPath(project)
+
+    // Avoid downloading multiple times
+    if (!validatorJarPath.toFile().exists()) {
```
yschimke, line 58: This is susceptible to corruption or partial downloads.  You might want to delete the file if partially created?

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/AdbInstall.kt
```diff
@@ -0,0 +1,66 @@
...
+            artifactLoader.load(apkLocation.get()) ?: throw GradleException("Cannot load APKs")
+        if (artifacts.elements.size != 1)
+            throw GradleException("Expected only one APK!")
+        val apkPath = File(artifacts.elements.single().outputFile).toPath()
+
+        val deviceArg = if (device.isEmpty()) "" else "-s $device "
+        val adbCommand = deviceArg + SET_WATCH_FACE_CMD.format(artifacts.applicationId)
+
+        project.exec {
+            it.commandLine("adb")
```
yschimke, line 56: Will this fail on multiple emulators/devices running? Does it need device also?

yschimke (approved):
Looks great.

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorDownload.kt
```diff
@@ -0,0 +1,70 @@
...
+import io.ktor.utils.io.ByteReadChannel
+import io.ktor.utils.io.core.isEmpty
+import io.ktor.utils.io.core.readBytes
+import kotlinx.coroutines.runBlocking
+import org.gradle.api.Project
+import java.nio.file.Files
+import java.nio.file.Path
+
+private const val VALIDATOR_URL =
+    "https://github.com/google/watchface/releases/download/release/dwf-format-2-validator-1.0.jar"
```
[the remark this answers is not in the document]
yschimke: I suspect if you make this a Task, then you can use build caching and not have to manage the file yourself.

https://docs.gradle.org/current/userguide/build_cache.html

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorExec.kt
garanj: Yeah, I did this because the `/*` in the path mucked up the comment, so it was invalid syntax. I'll try something else.

garanj: This works: `res/raw[-qualifier]/*.xml` ?

WatchFaceFormat/validator-plugin/settings.gradle.kts
yschimke: I think so. But really wanted the plugin to use a version catalog.

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorDownload.kt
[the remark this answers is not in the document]
garanj: Have moved to an input. Need to work out how to add version to TOML

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/ValidatorExec.kt
garanj: Thanks, have switched to a task with incremental changes

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/AdbInstall.kt
garanj: Yes, with multiple emulators, the task fails, so the user needs to specify e.g.

```
./gradlew validateWffAndInstall --device 12345678
```

garanj:
Thanks @yschimke I think I've addressed a bunch, but not all, of the comments.

I've shifted both the validator download and the WFF validation itself to being task-based, as suggested, including the latter being incremental, so getting the benefits of Gradle! (lots of learning there! :)

WatchFaceFormat/validator-plugin/plugins/src/main/kotlin/AdbInstall.kt
garanj: Oh, I missed this.

garanj: Fixed

yschimke:
@garanj none are blocking, feel free to land and iterate. But I do think it's worth ending up with a simple clean impl that works correctly with caching etc.

garanj:
> @garanj none are blocking, feel free to land and iterate. But I do think it's worth ending up with a simple clean impl that works correctly with caching etc.

Thanks! Will do. Caching should work fine now, using the inbuilt Gradle approach -- thanks for the guidance on this. Still some issues to resolve around TOML + libs.

PR #1207, opened 2024-11-11
Update Robolectric libraries to fix tests

kul3r4:
- Update dependency to Horologist and Robolectric to fix tests
- Update compile target to 35


kul3r4:
@yschimke these should be ready to go now

yschimke (approved):

PR #1241, opened 2025-01-10
Update kotlin in Complications sample

kul3r4:
Update kotlin and kotlinx versions in Complications to fix https://github.com/android/wear-os-samples/pull/1232 and some refactoring in the build.gradle


yschimke (approved):
Woot!

PR #1254, opened 2025-02-19
Fixes kotlin version compatibility

garanj:
Fixes kotlin compatibility issue when building watch faces validator plugin


yschimke (approved):
Can we update renovate to pin this dependency?

yschimke:
I guess either we stop running for this

https://github.com/android/health-samples/blob/main/renovate.json#L9

Or fileFilter?

https://docs.renovatebot.com/configuration-options/#filefilters

PR #1255, opened 2025-02-20
Ignores renovate rule for Kotlin in the case of Watch Face Format samples

garanj:
Does this configuration cover it?


yschimke (approved):
Yes, but I wonder if we jumped Kotlin above 2 anyway?

PR #1286, opened 2025-06-09
Update deps to Tiles 1.5/ProtoLayout 1.3

ithinkihaveacat:
Now stable:

  - https://developer.android.com/jetpack/androidx/releases/wear-tiles#1.5.0
  - https://developer.android.com/jetpack/androidx/releases/wear-protolayout#1.3.0

### DIFF
--- a/WearTilesKotlin/gradle/libs.versions.toml
+++ b/WearTilesKotlin/gradle/libs.versions.toml
@@ -1,12 +1,12 @@
 [versions]
-android-gradle-plugin = "8.10.0"
+android-gradle-plugin = "8.10.1"
 compose = "1.8.2"
 compose-compiler = "1.5.15"
 ktlint = "0.50.0"
 ktfmt = "0.54"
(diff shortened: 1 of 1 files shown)


WearTilesKotlin/gradle/libs.versions.toml
```diff
@@ -1,12 +1,12 @@
 [versions]
-android-gradle-plugin = "8.10.0"
+android-gradle-plugin = "8.10.1"
 compose = "1.8.2"
 compose-compiler = "1.5.15"
```
yschimke, line 4: Get the compose compiler through Kotlin 2?

https://kotlinlang.org/docs/compose-compiler-migration-guide.html

ithinkihaveacat:
@gemini-code-assist Re "The AGP version 8.10.1 does not seem to correspond to a standard public release": AGP 8.10.1 has been released: https://developer.android.com/build/releases/gradle-plugin#android-gradle-plugin-8.10.1

PR #1325, opened 2025-08-29
Datalayer improvement

ithinkihaveacat:
Try to make a few things clearer:

1. Values need to be retrieved synchronously.
2. You only get a value if it changes.


ithinkihaveacat:
> Can you show what is changing in the UI of the sample? I only see the log showing this value, how does it make it clear if this variables are not used anywhere? You could also add some comment explaining this new code.

Yes good point; added a comment since none of this appears in the UI (the existing response code isn't reflected in the UI either).

There is some separate code which updates the UI when a new count is received, however it's somewhat separate, and also doesn't show the actual count!

https://github.com/android/wear-os-samples/blob/b51a8c1f0c488c3f146eb03db5265c476f28e86c/DataLayer/Wearable/src/main/java/com/example/android/wearable/datalayer/MainViewModel.kt#L82-L86 

So there's quite a few ways this sample could be improved but just starting with this since my misunderstanding of this code cost me a bit of time…

PR #1335, opened 2025-09-24
Update Compose starter with more colors and additional button

kul3r4:
In an effort of showing more Material 3 features, such as more colors:
- Updated Theme using Material theme generator based on the existing primary color
- Used Tertiary 
- Added shape morphing effect to a button
- Added Button group
- Updated screenshots which weren't updated when we migrated to Material 3 for a roborazzi failure we didn't have time to investigate


ComposeStarter/app/src/main/java/com/example/android/wearable/composestarter/presentation/MainActivity.kt
```diff
@@ -198,26 +211,48 @@ fun ListScreen(modifier: Modifier = Modifier) {
...
-                    }
                         .transformedHeight(this, transformationSpec)
                 ) {
-                    Icon(
-                        imageVector = Icons.Rounded.Build,
-                        contentDescription = stringResource(
-                            R.string.example_button_content_description
+                    FilledIconButton(
+                        onClick = { showDialog = true },
+                        colors = IconButtonDefaults.iconButtonColors(
```
yschimke, line 232: Should this be filledIconButtonColors?

PR #1363, opened 2026-04-09
Update to navigation 3 and update padding

kul3r4:
Applies changes to use Compose 1.6 new features:
- navigation 3 for Wear OS
- enhanced padding support (it is possible now to remove Horologist layout)


ithinkihaveacat (approved):
LGTM

PR #1366, opened 2026-04-20
Remove Horologist from ComposeStarter

ithinkihaveacat:
## Summary / Goal

This PR removes the Horologist dependency from the `ComposeStarter` application
and migrates its screenshot tests to use Roborazzi directly.

## Motivation

The `ComposeStarter` app relied on Horologist primarily for its screenshot
testing infrastructure (`WearScreenshotTest` and `WearDevice`). To reduce
external dependencies and simplify the project, we are inlining the minimal
necessary code to handle screenshot qualifiers and Roborazzi interaction,
removing the need for Horologist in this module.

### [ComposeStarter]

- **Dependencies**:
  - Removed `horologist` version and `horologist-roboscreenshots` from
    `gradle/libs.versions.toml`.
  - Removed `testImplementation(libs.horologist.roboscreenshots)` from
    `app/build.gradle.kts`.

- **Testing Infrastructure**:
  - Added `WearScreenshotTest.kt` and `WearDevice.kt` to
    `src/test/java/presentation/` to replace the Horologist implementations.

- **Tests**:
  - Updated `GreetingScreenTest.kt`, `ListScreenTest.kt`, and
    `SampleDialogTest.kt` to remove Horologist imports and use the new local
    `WearScreenshotTest` base class.
  - Recorded new screenshots for all tests to match the new harness.

## Rationale & Safety

This change is safe as it only affects test code and dependencies. The
production code in `MainActivity.kt` is not modified. The new
`WearScreenshotTest` harness uses the same device qualifiers and Roborazzi
configuration as Horologist did, ensuring continuity in testing.

### Automated Tests

- Executed `./gradlew :app:testDebugUnitTest` in the `ComposeStarter` directory.
- Result: **BUILD SUCCESSFUL**. All tests passed.

### Visual Diff Analysis

We used an AI tool to compare the new screenshots against the originals from
`main`:

- **Layout**: The layout of central content and buttons remains identical.
- **Differences**: The time text ("10:10") at the top changed from white to gray
  (due to switching to native Material3 `TimeText` defaults), and button sizes
  in the dialog test were slightly reduced by ~10-15%. These are non-breaking
  changes consistent with the harness update.


kul3r4:
Looking at the screenshot tests images, it looks like with this change when you scroll it does not show the time text, is that expected?
Also, it looks like using com.google.android.horologist.screenshots.rng.WearDevice is still useful, instead of duplicating the code, maybe we keep the dependency only for testing given that list of devices is not officially maintained in DAC?

ithinkihaveacat:
> Looking at the screenshot tests images, it looks like with this change when you scroll it does not show the time text, is that expected? Also, it looks like using com.google.android.horologist.screenshots.rng.WearDevice is still useful, instead of duplicating the code, maybe we keep the dependency only for testing given that list of devices is not officially maintained in DAC?

@kul3r4  I believe this is expected. In Wear Compose Material 3,  ScreenScaffold  automatically hides the  TimeText  during scrolling to maximize content space. Horologist's AppScaffold didn't simulate this, but the native Material 3 components do, and so the screenshot tests now correctly reflect the actual behavior on a physical watch.

kul3r4:
@ithinkihaveacat Correct thank you for explaining!

ithinkihaveacat:
Thanks @gemini-code-assist your comments are all resolved now.

kul3r4 (approved):

PR #1367, opened 2026-04-24
Update to the 1.4/1.6 releases of android.wear.{protolayout,tiles}

ithinkihaveacat:
**Summary / Goal**
Update Wear OS Jetpack library dependencies to their final stable releases and migrate the main messaging tile to the modern API, while deferring changes to the reference layouts (Golden Tiles) to keep the PR focused.

**Motivation**
The project was previously using beta/RC versions of the `androidx.wear.tiles` and `androidx.wear.protolayout` libraries. With these libraries now stable, we can finalize the versions. The main sample has been migrated to `Material3TileService` natively to align with the latest best practices.

**Proposed Changes**
*   **Library Updates:** Updated `androidx-wear-tiles` to `1.6.0` and `androidx-wear-protolayout` to `1.4.0` in the version catalog.
    *   See: https://developer.android.com/jetpack/androidx/releases/wear-tiles#1.6.0
    *   See: https://developer.android.com/jetpack/androidx/releases/wear-protolayout#1.4.0
*   **Messaging Tile Migration:** Migrated `MessagingTileService` to extend `Material3TileService` and use the `tileResponse` method. Updated `Layout.kt` to use `basicImage` to avoid deprecation warnings associated with `setResourceId`.
*   **Excluded Golden Tiles:** The reference layouts in `com.example.wear.tiles.golden` have been intentionally excluded from this PR to simplify the review process. They still use the `BaseTileService` wrapper and rely on the Horologist dependency, which has been retained for now to support them. Issue to migrate: https://github.com/android/wear-os-samples/issues/1368

**Rationale & Safety**
This cleanup aligns the core sample with the intended stable usage of the APIs. Golden tiles remain functional on the old APIs and will be updated in a follow-up issue.

**Verification**
*   Performed a clean build (`./gradlew clean build`) successfully.
*   Verified that the tile renders correctly on the device.


WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -0,0 +1,90 @@
...
+        contacts = getMockNetworkContacts()
+
+        // For this sample, make Coil dumb by disabling all caching features.
+        SingletonImageLoader.setSafe { context ->
+            ImageLoader
+                .Builder(context)
+                .memoryCache(null)
+                .diskCache(null)
+                .build()
+        }
```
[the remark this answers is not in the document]
ithinkihaveacat: I kept the Coil configuration in the Service's  onCreate for this sample to minimize boilerplate and keep the code localized to the Tile implementation. While moving it to a custom  Application  class is definitely best practice for full applications, it felt excessive for the scope of this focused sample. I've noted this as a potential improvement for when the sample structure expands. Explained this rationale in a comment.

WearTilesKotlin/app/build.gradle.kts
```diff
@@ -47,24 +46,15 @@ android {
...
-    }
+
+
 }
 
-dependencies {
-    // Horologist provides helpful wrappers for Tiles development
-    implementation(libs.horologist.tiles)
 
+
```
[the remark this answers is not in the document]
ithinkihaveacat: Omitting the  kotlin  block is both necessary and correct here. With the upgrade to AGP 9.0, Kotlin support is now built-in, and the plugin manages the  kotlin  extension automatically. Explicitly adding the  kotlin  block or applying the  kotlin.android  plugin causes a direct conflict ('Cannot add extension with name kotlin') because the extension is already registered by the system. Since the project compiles successfully for all variants without it, and no missing opt-in errors are reported, leaving this block removed is the proper way to configure the build for this environment.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -0,0 +1,90 @@
...
+
+        imageLoader = SingletonImageLoader.get(this)
+    }
+
+    /** This method returns a Tile object, which describes the layout and resources of the Tile. */
+    override suspend fun MaterialScope.tileResponse(requestParams: TileRequest): Tile {
```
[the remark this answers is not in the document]
ithinkihaveacat: The suggestion to override  resourcesResponse  does not apply to  Material3TileService . In this new API,  onTileResourcesRequest  is  final  and designed as a no-op. Instead, resources are collected automatically when embedded in the layout using APIs like  basicImage . I verified that local resources (like the offline fallback) render correctly on the device using this approach.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -0,0 +1,94 @@
...
+        val layoutElement = tileLayout(contacts, imageResources)
+
+        val resourcesVersion = contacts.map { it.id }.toSortedSet().joinToString()
+
+        return Tile
+            .Builder()
+            .setResourcesVersion(resourcesVersion)
+            .setTileTimeline(Timeline.fromLayoutElement(layoutElement))
+            .build()
+    }
```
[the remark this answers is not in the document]
ithinkihaveacat: The suggestion to override  resourcesResponse  does not apply to  Material3TileService . In this new API,  onTileResourcesRequest  is  final  and designed as a no-op. Instead, resources are collected automatically when embedded in the layout using APIs like  basicImage . I verified that local resources (like the offline fallback) render correctly on the device using this approach.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -0,0 +1,94 @@
...
+                                id to image
+                            }
+                        }
+                    }.awaitAll()
+                    .toMap()
+            }
+
+        val layoutElement = tileLayout(contacts, imageResources)
+
+        val resourcesVersion = contacts.map { it.id }.toSortedSet().joinToString()
```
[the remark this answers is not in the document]
ithinkihaveacat: I kept the  resourcesVersion  tied strictly to the set of contact IDs intentionally. The previous implementation used random UUIDs for cache busting, but there was a risk that this was used in production code. I have added a comment in the source code noting that images could be cached by the system because of this.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/Contact.kt
```diff
@@ -44,14 +48,14 @@ fun getMockNetworkContacts() =
             id = 0,
             initials = "AC",
             name = "Ali C",
-            avatarSource = AvatarSource.Network("$avatarPath/ali.png")
+            avatarSource = AvatarSource.Network("$AVATAR_PATH/ali.jpg")
```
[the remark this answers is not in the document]
ithinkihaveacat: JPG is correct.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -0,0 +1,95 @@
...
+
+        // For this sample, make Coil dumb by disabling all caching features.
+        // Configured here for simplicity in this sample, though Application level is preferred in production.
+        SingletonImageLoader.setSafe { context ->
+            ImageLoader
+                .Builder(context)
+                .memoryCache(null)
+                .diskCache(null)
+                .build()
+        }
```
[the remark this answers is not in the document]
ithinkihaveacat: Keeping here for simplicity. There's a warning comment just above this:

> Configured here for simplicity in this sample, though Application level is preferred in production.

ithinkihaveacat:
FYI @NedaTop

kul3r4 (approved):

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/Layout.kt
```diff
@@ -38,27 +37,31 @@ import androidx.wear.protolayout.modifiers.LayoutModifier
...
-            setWidth(expand())
-            setModifiers(LayoutModifier.clip(shapes.full).toProtoLayoutModifiers())
-            setResourceId(contact.imageResourceId())
-            setContentScaleMode(CONTENT_SCALE_MODE_CROP)
-        }
+fun MaterialScope.contactButton(
+    contact: Contact,
+    imageResource: ResourceBuilders.ImageResource?
+): LayoutElement {
+    if (imageResource != null) {
```
NedaTop, line 56: Nit: it can be something like this below, but up to you:

```
imageResource?.let {
  protoLayoutScope.basicImage(...)
} ?: ...
```

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/Layout.kt
```diff
@@ -172,26 +182,59 @@ internal fun socialPreview5(context: Context) = socialPreviewN(context, 5)
...
+        onTileRequest = { request ->
+            TilePreviewHelper
+                .singleTimelineEntryTileBuilder(
+                    materialScopeWithResources(
+                        context = context,
+                        protoLayoutScope = request.scope,
+                        deviceConfiguration = request.deviceConfiguration,
+                        allowDynamicTheme = true,
+                        defaultColorScheme =
+                            androidx.wear.protolayout.material3
```
NedaTop, line 211: Nit: can you import directly?

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/Layout.kt
```diff
@@ -172,26 +182,59 @@ internal fun socialPreview5(context: Context) = socialPreviewN(context, 5)
...
+                                .ColorScheme()
+                    ) {
+                        tileLayout(contacts, imageResources)
+                    }
+                ).build()
+        },
+        onTileResourceRequest = { request ->
```
NedaTop, line 218: You shouldn't need this with the new resource handling. If there are some issues in not using it, ping me on chat to see

NedaTop: See here: https://developer.android.com/reference/androidx/wear/tiles/tooling/preview/TilePreviewData#TilePreviewData(kotlin.Function1,androidx.wear.protolayout.expression.PlatformDataValues,kotlin.Function1)

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -0,0 +1,95 @@
...
+
+        imageLoader = SingletonImageLoader.get(this)
+    }
+
+    /** This method returns a Tile object, which describes the layout and resources of the Tile. */
+    override suspend fun MaterialScope.tileResponse(requestParams: TileRequest): Tile {
+        // Asynchronously load images associated with resourceIds, and create a Map<String,
```
NedaTop, line 65: Do you need it to still be a String? Is ID in contacts String or it was a must from before?

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -0,0 +1,95 @@
...
+        // ResourceBuilders.ImageResource> suitable for adding to the Resources object
+        val imageResources =
```
NedaTop, line 67: Explicitly adding a type to this one would help with readability

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
```diff
@@ -0,0 +1,95 @@
...
+                                id to image
+                            }
+                        }
+                    }.awaitAll()
+                    .toMap()
+            }
+
+        val layoutElement = tileLayout(contacts, imageResources)
+
+        // Note: Images could be cached by the system because version is tied to contact IDs.
```
NedaTop, line 86: Not sure I get what is the meaning here, but with the new resource handling you can also omit the resource version, system adds it automatically as the hash of resources, so when there's a change it will know to re-fetch (reload) them

NedaTop (commented):
Thank you!

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/Layout.kt
ithinkihaveacat: Slightly tricky but why not?! Changed.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/Layout.kt
ithinkihaveacat: Updated to import.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
ithinkihaveacat: Added Map<String, ResourceBuilders.ImageResource?>.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
ithinkihaveacat: Done.

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/MessagingTileService.kt
ithinkihaveacat: Yes,  Contact.id  is a  Long , but we need a  String  here because the ProtoLayout API ( basicImage  and  addIdToImageMapping ) expects a string for resource IDs. If we only displayed local resources we could do away with this map entirely but it's needed to save key/value pairs for network images (which do not have the Android resource id we could otherwise use).

WearTilesKotlin/app/src/main/java/com/example/wear/tiles/messaging/Layout.kt
ithinkihaveacat: Removed.

ithinkihaveacat:
@NedaTop Thanks! I've made the changes you suggested with the exception of https://github.com/android/wear-os-samples/pull/1367#discussion_r3148886030 (which I think is okay as is). Can you take another pass?

NedaTop:
LGTM, thank you!

PR #1379, opened 2026-06-15
Add gesture support

ithinkihaveacat:


ithinkihaveacat:
Abandoned; we're doing this in snippets instead.

PR #1382, opened 2026-07-10
Update WearWidget sample dependencies and cleanup previews

yschimke:
- Update remote-compose to 1.0.0-alpha14
- Update remote-material3 to 1.0.0-alpha07
- Update glance-wear to 1.0.0-alpha13
- Remove obsolete preview snapshots and use official library previews instead
- Use remote-material3 RemoteText component
- Remove triggerUpdate RestrictedApi suppressions
- Remove file-level RestrictedApi suppressions as they are no longer needed after switching to non-restricted RemoteText


ithinkihaveacat:
Hey @yschimke did you manage to get the previews working in Android Studio? I had a similar PR (#1381) that I was holding off on merging because previews were broken due to the `Read access not allowed during rendering (/macros)` issue.

yschimke:
I'll hold off then.

